Issue #29, opened 2026-05-18
Add session support


jesperpedersen:
Use a session concept instead of global history.

Sessions are kept under ~/.orangu/sessions/ with a UUID identifier that is logged upon /quit

Keep the global history though and allow for the global history to be pulled in with a /history tool

jesperpedersen:
Add a `--resume` to orangu to choose a session

tejassinghbhati:
I am working on opening a PR on this, implementing UUID per session. I'd be thankful to be assigned.

jesperpedersen:
@tejassinghbhati Go for it !

jesperpedersen:
@tejassinghbhati You will need

```
--resume <UUID>
```

and update `/usage`

PR #37, opened 2026-05-20
feat(#29): add session support with UUID and resume

tejassinghbhati:
Each run gets a UUID stored at ~/.orangu/sessions/<uuid>/. Sessions store per-session readline history and chat messages. On /quit the UUID is printed so the user can resume. --resume <uuid> restores the previous conversation and history. /history shows the global command history. Global ~/.orangu/orangu.history continues to be appended.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1280,6 +1280,7 @@ dependencies = [
  "terminal_size",
  "tiktoken-rs",
  "tokio",
+ "uuid",
  "walkdir",
 ]
 
@@ -2470,6 +2471,17 @@ name = "utf8parse"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
+
+[[package]]
+name = "uuid"
+version = "1.23.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddd74a9687298c6858e9b88ec8935ec45d22e8fd5e6394fa1bd4e99a87789c76"
+dependencies = [
+ "getrandom 0.4.2",
+ "js-sys",
+ "wasm-bindgen",
+]
 
 [[package]]
 name = "version_check"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ syntect = "5.3.0"
 terminal_size = "0.4.1"
 tiktoken-rs = "0.6.0"
 tokio = { version = "1.48.0", features = ["full"] }
+uuid = { version = "1", features = ["v4"] }
 walkdir = "2.5.0"
 
 [dev-dependencies]
--- a/doc/manual/en/03-quickstart.md
+++ b/doc/manual/en/03-quickstart.md
@@ -79,4 +79,12 @@ Then start with:
 
 By default the tools operate on the current directory. Use `--workspace /path/to/project` to point **orangu** at another tree.
 
+Each run creates a session. When you exit, the resume command is printed:
+
+```text
+orangu --resume 550e8400-e29b-41d4-a716-446655440000
+```
+
+Use `/sessions` to list previous sessions and find a UUID to resume.
+
 Lines whose first non-whitespace character is `#` stay local and are not sent to the model. Lines whose first non-whitespace character is `\` are ignored.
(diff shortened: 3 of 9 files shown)


jesperpedersen:
@tejassinghbhati Remember your `Squash`

tejassinghbhati:
I have added --resume <UUID> to the /usage output so you can grab it directly from there without waiting for /quit.

jesperpedersen:
```
/usage
...
Resume           : --resume 3663b5ba-da6d-4d67-81c1-fe18a396cb38
```

to

```
/usage
...
Session          : 3663b5ba-da6d-4d67-81c1-fe18a396cb38
```

jesperpedersen:
@tejassinghbhati Maybe it should just say

```
orangu --resume 3663b5ba-da6d-4d67-81c1-fe18a396cb38
```

after /quit

jesperpedersen:
@tejassinghbhati We likely need a /session command ... Thinking about it ... Do we really need /history ? Because we have <ARROW_UP> + <ARROW_DOWN> ...

jesperpedersen:
@tejassinghbhati I think <ARROW_UP> + <ARROW_DOWN> is enough within  the session. We can remove the global history

jesperpedersen:
@tejassinghbhati So, no /session and /history - it is too confusing

tejassinghbhati:
On the global history, I kept the file writes since arrow up/down loads from it on startup. Removing it would mean each new session starts blank, which felt like a regression. The /history command that displayed it is gone though. Let me know if you'd prefer it fully removed.

tejassinghbhati:
If you want to use a session concept and not history so the change would be: on startup, load history from the session file only (not the global file), and stop writing to the global file entirely. Each session gets its own isolated history.

jesperpedersen:
@tejassinghbhati Yes, we can remove global history and keep it at session level. So, yes, each new session start fresh

tejassinghbhati:
Yeah, so most of the changes have been done now, I believe history now lives only at the session level, so each new session starts fresh. Also fixed the remaining two quit paths to print orangu --resume <uuid> and updated the manual accordingly.

jesperpedersen:
@tejassinghbhati Great ! Now, we need the `/sessions` tool that will display all known session UUIDs and their stats. Maybe

```
<UUID>   #NumberOfCommands   <FIRST_COMMAND>
```

Maybe we should stick a `metadata` file in each session that records the start time, and last updated. Then we can display that in `/sessions` as well.

Play around with it

jesperpedersen:
@tejassinghbhati We should likely have the workspace in the `metadata`

tejassinghbhati:
sure, will add a metadata file per session (start time, last updated, workspace) and a /sessions command to list them.

jesperpedersen:
@tejassinghbhati We should probably sort in `/sessions` them based on the `CREATED_AT` timestamp

jesperpedersen:
@tejassinghbhati You can do optional filters too - so only sessions for a specific workspace are shown

tejassinghbhati (commented):
Session lifecycle (new/resume/quit), metadata, and /sessions filtering all work correctly and consistently. One minor bug: the ... truncation check in list_sessions_output compares byte length vs char count — first_cmd.len() > truncated.len() should be first_cmd.chars().count() > 60.

jesperpedersen:
@tejassinghbhati Just stick it on a single line - aligned as columns - as short as possible. And, upon /quit show the resume needed in order to get back into the same session

tejassinghbhati (commented):
Session lifecycle is consistent across all three quit paths, metadata and message persistence are straightforward, and the /sessions columnar output is now exactly how it had to be done. 

The only remaining nit is that load_history is called just for .len(), it loads the full history into memory and discards it, which is wasteful for large histories but not a real problem at this scale.

jesperpedersen:
@tejassinghbhati Add a LAST column too. Also, after /quit I'm missing the resume command in the terminal

```
/quit
orangu --resume <UUID>
```

jesperpedersen:
@tejassinghbhati You probably need a time too to `START` and `LAST`, and you can make them short like

```
202605220914
```

tejassinghbhati (commented):
There is consistency with the session lifecycle now. Also now the columnar /sessions output with compact timestamps. 
The drop(_terminal_ui_guard) before the resume eprintln! is the right fix, it prints to the primary screen after leaving the alternate buffer. No issues.

jesperpedersen:
@tejassinghbhati We need a bit more documentation for this, otherwise it looks good

jesperpedersen:
Merged.

jesperpedersen:
@tejassinghbhati We have to think about switch session after startup, and remove "empty" sessions

tejassinghbhati:
I am surely looking forward on working on both the follow ups.

Issue #30, opened 2026-05-18
Add workspace support


jesperpedersen:
Add internal tabs for workspaces with shortcuts.

Each workspace is its own session.

This is instead of starting multiple orangu instances

jesperpedersen:
~~Maybe we need a shortcut "area" - so "Alt+x -> 1" is Workspace 1 - for example~~

jesperpedersen:
> Switching to a workspace should automatically switch branch if multiple workspaces are working on the same repo

I don't know if we should do this to begin with

tejassinghbhati:
I tried breaking the whole thing into smaller concrete steps so the rollout feels incremental instead of trying to solve the entire session UX in one go.

### Immediate (this PR)

- Add branch to `SessionMetadata` so workspace + branch effectively become the natural session identity
- `/session <TAB>` cycles through session UUIDs newest-to-oldest
- Empty session cleanup on exit:
  - delete only if there were zero LLM queries
  - only on `main` / `master`
  - feature branch sessions always persist

### Config (`orangu.conf`)

- Add `auto_resume = true`
- When enabled, startup auto-detects the current workspace + branch and silently restores the matching session

### Later

- Interactive picker when multiple sessions match the same branch
- Expand `/session` into a full mid-run session switcher

I’m starting with the branch metadata + tab completion + cleanup logic first since those pieces are tightly connected and they unblock the rest of the workflow pretty naturally.

jesperpedersen:
We likely need an indicator (🟢 / 🔴) for each workspace if it is still valid or not

jesperpedersen:
@tejassinghbhati Sessions are separate from workspaces - workspaces build upon sessions

jesperpedersen:
We can likely use the NumLock keys to switch workspaces

jesperpedersen:
@tejassinghbhati I'm thinking `Alt+,` (Move to left workspace) - `Alt+.` (Move to right workspace), `Alt+n` (New workspace on right), `Alt+x` (Close workspace)

jesperpedersen:
@tejassinghbhati We probably need 

```
/workspace <TAB>
```

which will `<TAB>` complete on directories based on path given

jesperpedersen:
@tejassinghbhati Of course we have to use a line under the status bar to indicate the workspace and its status

```
------------
**1** 🟢 | 2 🔴 
------------ 
```

where bold shows the active workspace

jesperpedersen:
[...]

PR #40, opened 2026-05-23
feat(#30): branch-aware auto-resume, /session <TAB>, and empty session cleanup

tejassinghbhati:
- Add `branch: Option<String>` to SessionMetadata (#[serde(default)] for backward compat)
- On startup, auto-detect workspace + branch and silently resume the matching session if exactly one non-empty match exists; print "Resuming session <uuid>" before TUI
- On exit, delete session directory if zero LLM queries and on main/master/no-branch (feature branch sessions always persist); skip resume command print when deleted
- Add /session [uuid] command: no arg lists sessions, uuid arg prints resume command
- Add /session <TAB> completion cycling session UUIDs newest-to-oldest via mtime
- Add BRANCH column to /sessions output

### DIFF
--- a/doc/manual/en/03-quickstart.md
+++ b/doc/manual/en/03-quickstart.md
@@ -76,16 +76,18 @@ Then start with:
 /squash
 /delete feature/foo
 /sessions
-/session 550e8400-e29b-41d4-a716-446655440000
+/session <UUID>
 ```
 
 By default the tools operate on the current directory. Use `--workspace /path/to/project` to point **orangu** at another tree.
 
-**orangu** automatically resumes an existing session when you return to the same workspace and Git branch. When a previous session is found, the UUID is printed before the TUI starts:
+**orangu** automatically resumes an existing session when you return to the same workspace and Git branch. When a previous session is found, the status bar shows:
 
 ```text
 Resuming session 550e8400-e29b-41d4-a716-446655440000
 ```
+
+for five seconds or until the first command is run.
 
 On exit, the resume command is printed so you can return to the session from a different branch or machine:
 
--- a/doc/manual/en/03-quickstart.md
+++ b/doc/manual/en/03-quickstart.md
@@ -75,16 +75,26 @@ Then start with:
 /init_repo
 /squash
 /delete feature/foo
+/sessions
+/session 550e8400-e29b-41d4-a716-446655440000
 ```
 
 By default the tools operate on the current directory. Use `--workspace /path/to/project` to point **orangu** at another tree.
 
-Each run creates a session. When you exit, the resume command is printed:
+**orangu** automatically resumes an existing session when you return to the same workspace and Git branch. When a previous session is found, the UUID is printed before the TUI starts:
+
+```text
+Resuming session 550e8400-e29b-41d4-a716-446655440000
+```
+
+On exit, the resume command is printed so you can return to the session from a different branch or machine:
 
 ```text
 orangu --resume 550e8400-e29b-41d4-a716-446655440000
 ```
 
-Use `/sessions` to list previous sessions and find a UUID to resume.
+Sessions that had no LLM interaction on `main`, `master`, or outside a Git repository are deleted automatically on exit. Feature branch sessions are always kept.
+
+Use `/sessions` to list all sessions and their branches. Use `/session <uuid>` and Tab completion to find and print the resume command for a specific session.
 
 Lines whose first non-whitespace character is `#` stay local and are not sent to the model. Lines whose first non-whitespace character is `\` are ignored.
--- a/doc/manual/en/40-terminal.md
+++ b/doc/manual/en/40-terminal.md
@@ -48,13 +48,13 @@ Each run of `orangu` creates or resumes a session identified by a UUID.
 
 ### Automatic resume
 
-On startup, `orangu` checks whether a session already exists for the current workspace path and Git branch. If exactly one matching session with conversation history is found, it is resumed automatically:
+On startup, `orangu` checks whether a session already exists for the current workspace path and Git branch. If exactly one matching session with conversation history is found, it is resumed automatically. The status bar lower-left shows:
 
 ```text
 Resuming session 550e8400-e29b-41d4-a716-446655440000
 ```
 
-This message appears on the primary terminal before the TUI starts. No `--resume` flag is needed for normal branch-based workflows.
+for five seconds or until the first command is run. No `--resume` flag is needed for normal branch-based workflows.
 
 If more than one session matches the current workspace and branch, a fresh session is started instead. Use `--resume <uuid>` to target a specific session explicitly.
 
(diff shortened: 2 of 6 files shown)


jesperpedersen:
@tejassinghbhati We need a list of branches since you can work on multiple branches within a repository in a single session

tejassinghbhati:
Sure, I will implement that. Just need some time to fix a few things

tejassinghbhati (commented):
### `commands.rs`

Clean addition of `Session(Option<Cow<'a, str>>)` alongside the existing `Sessions` variant.

Parsing follows the existing codebase pattern consistently:

- exact-match handling in the main `match` block
- with-args handling in the `_ =>` fallback
- natural-language aliases wired through `matches_ci`

No issues here.

---

### `completion.rs`

`session_uuids_newest_first` reads `~/.orangu/sessions/`, converts filesystem `mtime` into `u64` seconds, sorts using `sort_by_key(Reverse)`, and returns UUIDs newest-to-oldest.

Nicely integrated into `completion_candidates` before the generic `/` prefix handler, so `/session <TAB>` correctly receives UUID completions.

Graceful degradation is handled well:

- missing sessions directory returns an empty list
- unreadable entries are skipped silently

Clean implementation overall.

---

### `tui.rs`

Single-line addition to the help text.

Correctly positioned and clearly mentions Tab completion behaviour.

Looks good.

---

### `main.rs`

This is where most of the substantive logic lives.

A few implementation notes:

- `find_session_for_workspace_branch` correctly uses:

  ```rust
  let Some(...) = ... else { continue };
  ```

  which ensures malformed entries are skipped without aborting the entire scan.

- The `candidates.len() == 1` guard is the right conservative behaviour. Ambiguity falling back to a fresh session is the safest default.

- `has_messages` using `m.len() > 2` is a practical and reasonable non-empty check (`[]` serializes to 2 bytes).

- `is_ephemeral_branch` implemented with `matches!` is concise and readable.

- `delete_session_dir` being best-effort and silent is correct for cleanup paths.

- `eprintln!("Resuming session …")` happens before `TerminalUiGuard::new()`, so it appears on the primary screen rather than inside the alternate screen buffer — same correct placement as the earlier PR fix.

- `branch: Option<String>` combined with `#[serde(default)]` keeps old metadata files compatible without requiring migration logic.

One detail worth explicitly noting:

```rust
std::fs::read_dir(...).ok()?
```

inside `find_session_for_workspace_branch` will return `None` for the entire function if the sessions directory itself cannot be read, which is fine and effectively means “no sessions found.” Everything inside the iteration loop uses the correct early-continue pattern afterward.

---

Overall this looks solid:

- logically sound
- clippy-clean
- backward-compatible
- implementation style matches the surrounding codebase well

jesperpedersen:
@tejassinghbhati You need docs

tejassinghbhati:
The three features are pretty clean now,
Auto-resume: workspace+branch lookup with a conservative ambiguity fallback
Empty session cleanup: single epilogue guard, right logic
/session <TAB>: u64 mtime sort sidesteps the SystemTime: !Copy clippy issue

also the docs are complete

tejassinghbhati (commented):
Docs cover all three features with accurate examples and the updated metadata schema. Nothing missing.

doc/manual/en/03-quickstart.md
```diff
@@ -75,16 +75,26 @@ Then start with:
...
+/session 550e8400-e29b-41d4-a716-446655440000
 ```
 
 By default the tools operate on the current directory. Use `--workspace /path/to/project` to point **orangu** at another tree.
 
-Each run creates a session. When you exit, the resume command is printed:
+**orangu** automatically resumes an existing session when you return to the same workspace and Git branch. When a previous session is found, the UUID is printed before the TUI starts:
+
+```text
+Resuming session 550e8400-e29b-41d4-a716-446655440000
```
jesperpedersen, line 87: I think we can stick this inside the UI -- in the status bar, lower-left -- for 5 seconds or until the first command is run

doc/manual/en/03-quickstart.md
```diff
@@ -75,16 +75,26 @@ Then start with:
...
+On exit, the resume command is printed so you can return to the session from a different branch or machine:
 
 ```text
 orangu --resume 550e8400-e29b-41d4-a716-446655440000
 ```
 
-Use `/sessions` to list previous sessions and find a UUID to resume.
+Sessions that had no LLM interaction on `main`, `master`, or outside a Git repository are deleted automatically on exit. Feature branch sessions are always kept.
+
+Use `/sessions` to list all sessions and their branches. Use `/session <uuid>` and Tab completion to find and print the resume command for a specific session.
```
jesperpedersen, line 100: Use `<UUID>`

doc/manual/en/40-terminal.md
```diff
@@ -44,12 +44,42 @@ Press `Esc` twice within 2 seconds during the waiting state to cancel the active
...
 
-Each run of `orangu` creates a unique session identified by a UUID. When the client exits, the full resume command is printed so you can resume later:
+Each run of `orangu` creates or resumes a session identified by a UUID.
+
+### Automatic resume
+
+On startup, `orangu` checks whether a session already exists for the current workspace path and Git branch. If exactly one matching session with conversation history is found, it is resumed automatically:
+
+```text
+Resuming session 550e8400-e29b-41d4-a716-446655440000
```
jesperpedersen, line 54: Same

doc/manual/en/40-terminal.md
```diff
@@ -60,40 +90,49 @@ metadata   session metadata (JSON)
...
   "started_at": 1748000000,
   "last_updated_at": 1748003600,
-  "workspace": "/home/user/myproject"
+  "workspace": "/home/user/myproject",
+  "branch": "feature/my-pr"
 }
 ```
 
-Timestamps are Unix seconds (UTC).
+`branch` is `null` for sessions started outside a Git repository or in a detached HEAD state. Timestamps are Unix seconds (UTC).
```
jesperpedersen: It should be empty string

jesperpedersen (changes requested):

doc/manual/en/03-quickstart.md
tejassinghbhati: Sure I did that, moved to the status bar. Shows as a StatusFragment on the lower-left for 5 seconds or until the first command clears it.

doc/manual/en/40-terminal.md
tejassinghbhati: Yes. docs now say the status bar shows the message rather than the primary terminal.

doc/manual/en/03-quickstart.md
tejassinghbhati: Fixed. used <UUID>

tejassinghbhati (commented):
- Changed branch field from Option<String>/null to String/"" (empty string
  for sessions outside a Git repo or in detached HEAD)
- Show auto-resume notice in the status bar (lower-left) for 5 seconds
  or until first command, instead of printing to the terminal pre-TUI
- Update docs: /session <UUID> placeholder, branch empty string wording,
  status bar location for the resume notice

doc/manual/en/40-terminal.md
tejassinghbhati: Changed. branch is now a String (serialises as "") instead of Option<String> (null). Updated is_ephemeral_branch and find_session_for_workspace_branch signatures accordingly, and fixed the docs to say "empty string".

jesperpedersen:
Merged.

PR #73, opened 2026-06-04
[#71] /grep: Search the workspace with git grep

### LINKED ISSUE #71: tool: /grep
Create a tool called `/grep` that will use `git grep` to search the workspace.

```
/grep delta
find delta
grep delta
```

tejassinghbhati:
Closes #71

Adds a `/grep <pattern>` command that searches all tracked files in the workspace using `git grep --line-number`.

**New `/grep` command:**
- `/grep <pattern>` — search tracked files for a pattern
- Natural-language: `grep <pattern>`, `find <pattern>`, `git grep <pattern>`
- Gracefully handles exit code 1 (no matches) with a friendly message
- Requires a Git repository; always uses plain Git (`gh` has no equivalent)

**Files changed:**
- `commands.rs` — `Grep` variant, slash + NL parsers, usage message
- `git.rs` — `grep_output` function
- `main.rs` — dispatch handler
- `tui.rs` — help text + NL forms sentence
- `completion.rs` — added `/grep` to command list
- `40-terminal.md` — command table, command notes, NL aliases

### DIFF
--- a/doc/manual/en/40-terminal.md
+++ b/doc/manual/en/40-terminal.md
@@ -179,6 +179,7 @@ All slash commands are handled locally. They are not sent to the model.
 | `/comment <number> "<comment>"` | Add a comment to a GitHub issue with gh issue comment |
 | `/commit <message>` | Commit all tracked changes with git commit -a -m |
 | `/diff [branch]` | Show a color unified diff; without a branch shows unstaged changes, with a branch shows changes since diverging from it |
+| `/grep <pattern>` | Search the workspace for a pattern with git grep |
 | `/init_repo` | Initialize a Git repository in the workspace |
 | `/log` | Show commit log (uses `git lg` alias if configured) |
 | `/merge <branch>` | Merge a branch into the current branch |
@@ -210,6 +211,7 @@ Free-form prompts are blocked when the server or model status in the header is r
 - `/open_file <path>` is workspace-scoped; paths outside the workspace are rejected. It launches `$EDITOR` on the file in a separate window so orangu stays usable, and never waits for the editor.
 - `/show_file [--hash] [--author] <path> [<ref>]` is workspace-scoped; without a ref, the current workspace file is shown — when `bat` is installed it is used for the plain view, otherwise the built-in syntax-highlighted renderer is used; when a ref (commit hash, branch, or tag) is given, the file content at that ref is retrieved via `git show <ref>:<path>` and rendered with the built-in renderer; `--hash` and `--author` add per-line blame columns sourced from `git blame`, using the same ref when one is provided; Tab completion for the first positional argument offers workspace file paths recursively; Tab completion for the second positional argument cycles through that file's commit history (abbreviated hashes from `git log --follow`)
 - `/build` detects the project type from the workspace root and runs the appropriate toolchain: for Rust (`Cargo.toml`) it runs `cargo fmt`, `cargo clippy`, `cargo build`, and `cargo test`; for C (`CMakeLists.txt`) it runs `clang-format.sh` (if present), creates a `build/` directory if needed, runs `cmake ..` on the first build, then `make`; for Java (`pom.xml`) it installs frontend dependencies with `npm ci` when outdated, runs `npm run fix` and `npm run check` for the frontend (if `src/frontend/` exists), then `mvn package`; each step is reported individually and the pipeline stops on first failure
+- `/grep <pattern>` requires a Git repository and runs `git grep <pattern>` to search all tracked files; output is piped through the configured non-interactive pager (`pager.grep`, then `core.pager`) when one is set — if `delta` is configured it will colorize and format the results; exit code 1 (no matches) is handled gracefully; `gh` has no equivalent so it always uses plain Git; natural-language forms `grep <pattern>` and `find <pattern>` are also handled
 - `/diff` uses `git diff` inside Git repositories and applies configured non-interactive Git pagers such as `delta`; outside Git repositories it keeps the existing non-Git behavior; `/diff <branch>` runs `git diff <branch>...HEAD` to show commits on the current branch not yet in the specified branch; Tab completion after `/diff ` or natural-language forms such as `diff against <branch>` offers local and remote branch names
 - `/status` requires a Git repository and runs `git status --branch --short`; `gh` has no equivalent so it always uses plain Git; added files and untracked entries are shown in green, deleted entries in red, and modified entries in the default terminal color; the branch line is shown in a muted color
 - `/log` requires a Git repository; if a `lg` alias is found in `~/.gitconfig` it runs `git lg`, otherwise it falls back to `git log --graph --oneline --decorate`; see the optional tools chapter for the recommended `git lg` alias setup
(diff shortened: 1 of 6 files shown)


src/bin/orangu/git.rs
```diff
@@ -724,6 +724,32 @@ pub fn run_git_diff_pager(
     String::from_utf8(output.stdout).context("git pager output was not UTF-8")
 }
 
+pub fn grep_output(workspace: &Path, pattern: &str) -> Result<String> {
+    let repo_root = discover_git_root(workspace)
+        .ok_or_else(|| anyhow!("grep is only available inside a Git repository"))?;
+    let output = std::process::Command::new("git")
+        .arg("-C")
+        .arg(&repo_root)
+        .args(["grep", "--line-number", "--color=never", pattern])
```
jesperpedersen: I don't think this will work as people have their own configuration.

So remove `--line-number` and `--color=never`. Remember documentation as well

Maybe we can do overrides in `~/.orangu/orangu.conf` that have a high priority than `~/.gitconfig` - but that is a separate issue.

tejassinghbhati: sure, I have removed `--line-number` and `-color=never` so the output respects each user's own `git grep` configuration. Also I have accordingly updated the documentation. 

Yeah, we can surely look forward to implementing the config override idea.

tejassinghbhati (commented):
- **`commands.rs`** : `Grep` variant fits naturally into the `LocalCommand` enum; slash parser and
  NL parser (`grep`, `find`, `git grep` prefixes) are consistent with how every other command is
  handled. Usage message follows the same pattern.

- **`git.rs`** : `grep_output` is tight: Git-repo guard, graceful exit-code-1 handling (no matches
  isn't an error), and stderr surfaced on real failures. No hardcoded flags so user config is
  respected.

- **`main.rs`** : dispatch follows the exact same `None` → usage error / `Some` → run pattern as
  `/diff`, `/restore`, etc. Nothing surprising.

- **`tui.rs`** : help text entry sits in the right alphabetical slot (`/diff → /grep → /init_repo`);
  NL forms added to the inline sentence correctly.

- **`completion.rs`** : one-liner, `/grep` slotted in the right place.

- **`40-terminal.md`** : all three sections updated (command table, command notes, NL aliases) and
  consistent with each other.

this pr is small, correct, and follows every existing convention. No concerns.

jesperpedersen:
There is still something wrong. There are no colors inside orangu, but I have it on the console as I'm using delta. Maybe install delta and bat locally, and configure it. See 72-extra.md

doc/manual/en/40-terminal.md
```diff
@@ -209,6 +210,7 @@ Free-form prompts are blocked when the server or model status in the header is r
...
 - `/build` detects the project type from the workspace root and runs the appropriate toolchain: for Rust (`Cargo.toml`) it runs `cargo fmt`, `cargo clippy`, `cargo build`, and `cargo test`; for C (`CMakeLists.txt`) it runs `clang-format.sh` (if present), creates a `build/` directory if needed, runs `cmake ..` on the first build, then `make`; for Java (`pom.xml`) it installs frontend dependencies with `npm ci` when outdated, runs `npm run fix` and `npm run check` for the frontend (if `src/frontend/` exists), then `mvn package`; each step is reported individually and the pipeline stops on first failure
+- `/grep <pattern>` requires a Git repository and runs `git grep --color=always <pattern>` to search all tracked files; output is piped through the configured non-interactive pager (`pager.grep`, then `core.pager`) when one is set — if `delta` is configured it will colorize and format the results; exit code 1 (no matches) is handled gracefully; `gh` has no equivalent so it always uses plain Git; natural-language forms `grep <pattern>` and `find <pattern>` are also handled
```
jesperpedersen: Now, we have `--color=always`

src/bin/orangu/git.rs
```diff
@@ -724,6 +732,38 @@ pub fn run_git_diff_pager(
...
 }
 
+pub fn grep_output(workspace: &Path, pattern: &str) -> Result<String> {
+    let repo_root = discover_git_root(workspace)
+        .ok_or_else(|| anyhow!("grep is only available inside a Git repository"))?;
+    let terminal_width = current_terminal_width();
+    let output = std::process::Command::new("git")
+        .arg("-C")
+        .arg(&repo_root)
+        .args(["grep", "--color=always", pattern])
```
jesperpedersen: Here

doc/manual/en/40-terminal.md
tejassinghbhati: yeah, apologies for that! Removed --color=always from the doc as well — updated to just git grep <pattern>.

src/bin/orangu/git.rs
tejassinghbhati: fixed yep

jesperpedersen:
It seems like git diff doesn't pickup my ~/.gitconfig because the colors are gone from the delta and bat configuration again

tejassinghbhati:
Okay, yeah I had to rebase on upstream which included recently added changes. There was some problem since if delta (or bat) is installed via cargo and `~/.cargo/bin` is only added to PATH in `~/.bashrc`, the login shell won't find it. Changed to `sh -c` so the pager runs with the inherited PATH from the user's running session.

jesperpedersen:
@tejassinghbhati Did you install delta and bat ? I'm still not getting any colors inside orangu but I do in the terminal

tejassinghbhati:
yes I have installed delta and bat via scoop

jesperpedersen:
@tejassinghbhati And, do you have colors in the terminal with a `git grep` and none inside orangu ?

tejassinghbhati:
Configured in ~/.gitconfig with core.pager = delta, interactive.diffFilter = delta --color-only, and the standard delta settings (dark, side-by-side, line-numbers, navigate). The sh -lc → sh -c fix in this PR should also resolve the colors issue on your end, the login shell was dropping PATH entries like ~/.cargo/bin where delta lives. Once this merges you should get colors in orangu matching what you see in the terminal.

tejassinghbhati:
Yes, delta and bat are installed and configured. Here's orangu running on the branch, server and model are red since I'm not running a local LLM server, but the app itself is up. On Windows in PowerShell git grep doesn't auto-color without a TTY, so we likely need --color=always back, same as git diff --color=always — that's what gives delta colored input to work with when output is captured. Happy to add it back if you agree.
Screenshot 2026-06-04 221039

jesperpedersen:
@tejassinghbhati I have

```
/usr/bin/delta
/usr/sbin/bat
```

so maybe it is a Linux vs/ Windows thing

tejassinghbhati:
hmm. Adding --color=always back to the grep call (same as diff already does) might fix it for both platforms.

jesperpedersen:
@tejassinghbhati I don't think so as delta has a different theme than `--colors=always`

tejassinghbhati:
hmm, what do you think then would be the right approach here?

jesperpedersen:
@tejassinghbhati Let me do a deep review of this, and merge it. Then I can see if I can find the issue, and you can look at it on your end

jesperpedersen:
Rebased, and merged.

PR #96, opened 2026-06-09
[#95] feat: line_endings property + .gitattributes

### LINKED ISSUE #95: feat: line_endings - add line_endings property to orangu.conf and orangu --init
On Windows, git defaults to `core.autocrlf=true`, which auto-converts LF→CRLF on checkout. This causes `git status` (and orangu's `/status`) to show every file as modified even when nothing has changed, making it confusing to work with orangu on Windows.
Solution:
1. **`line_endings` property in `orangu.conf`** — default `unix`. orangu uses this when invoking git operations so line ending behaviour is explicit and 
   consistent across platforms.

2. **Support in `orangu --init`** — the interactive wizard should prompt for `line_endings` with auto-completion for the valid values (`unix`, `windows`, 
   `auto`).

3. **`.gitattributes`** — add `* text=auto eol=lf` at the repo root to enforce LF for all contributors regardless of local git config.

tejassinghbhati:
## Summary

- Add `line_endings` property to `[orangu]` section in `orangu.conf` (default: `unix`, valid values: `unix`, `windows`, `legacy_mac`)
- Add `line_endings` prompt to `orangu --init` wizard with TAB completion over the three valid values
- Add `.gitattributes` enforcing `eol=lf` for all contributors regardless of local git config

## Context

Windows contributors with `core.autocrlf=true` saw every file marked as modified in `/status` even when nothing had changed — git was auto-converting LF→CRLF on checkout. The `.gitattributes` file fixes this at the repo level, and the new config property makes the preferred line ending explicit and configurable per-workspace.

Closes #95

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


jesperpedersen:
@tejassinghbhati It isn't used though in the code ... was `.gitattributes` enough to fix your environment ?

tejassinghbhati:
image
@jesperpedersen Yes, `.gitattributes` alone fixed the environment, `/status` is clean now.

You're right that `line_endings` isn't wired up yet. What should it drive? My thinking was to pass `-c core.eol=<value>` to git commands when invoking them, as a fallback when no `.gitattributes` is present in the workspace. Does that match your intent, or did you have something else in mind?

jesperpedersen:
@tejassinghbhati Well, we are not a ~/.gitconfig editor so I think we will just leave it at ~/.gitattributes - so you can remove the rest for now

tejassinghbhati:
removed `line_endings` from config, init wizard, and the sample config. Only `.gitattributes` remains.

jesperpedersen:
Merged.

Issue #97, opened 2026-06-10
`install.sh` one-liner installer script


tejassinghbhati:
### Summary

Add an `install.sh` hosted at `https://mnemosyne-systems.github.io/orangu/install.sh` so users can install orangu with:

```sh
curl -fsSL https://mnemosyne-systems.github.io/orangu/install.sh | sh
```
### Requirements

- Auto-detect OS and architecture, pull the correct binary from `/releases/latest/`
- Install to `~/.local/bin` by default
- Check if `~/.local/bin` is on `$PATH` — if not, print a clear message telling the user how to add it
- Always pulls latest, no version pinning
- Shell completions deferred — depends on a `-s` flag being added to orangu itself first

### Platforms

| OS | Architecture |
|---|---|
| Linux | x86_64 (gnu) |
| Linux | x86_64 (musl) |
| Linux | aarch64 (gnu) |
| Linux | aarch64 (musl) |
| macOS | x86_64 |
| macOS | aarch64 (Apple Silicon) |

### Out of scope

- Windows (no `.sh` equivalent yet)
- Shell completions (tracked separately)

jesperpedersen:
@tejassinghbhati Can we do a `.cmd` for Windows ?

jesperpedersen:
@tejassinghbhati Maybe we should do `-s` first

PR #98, opened 2026-06-10
[#97] -s / --shell-completions flag

tejassinghbhati:
Auto-detects the current shell from $SHELL and prints the appropriate completion script to stdout. Supports bash, zsh, and fish.

  bash: eval "$(orangu -s)"
  zsh:  orangu -s > ~/.zsh/completions/_orangu
  fish: orangu -s > ~/.config/fish/completions/orangu.fish

Also updates all three contrib/shell completion files to include -s.


jesperpedersen:
@tejassinghbhati Lets bring the shell scripts inside orangu itself - maybe shell.rs. Then nuke contrib/shell and add documentation

jesperpedersen:
Merged.

PR #106, opened 2026-06-14
[#105] install.sh / install.cmd: one-liner installer for Linux, macOS and Windows

### LINKED ISSUE #105: `install.sh` / `install.cmd` one-liner installer
### Problem

There is currently no way to install orangu without building from source. A one-liner installer would significantly lower the barrier to entry, but a naive implementation has several pitfalls that need to be addressed properly.

### 1. libc detection on Linux

`uname -m` only gives the CPU architecture — it does not tell you whether the system uses glibc or musl. Picking the wrong target triple results in a binary that silently crashes or fails to link. The script needs to probe the system libc before selecting a release asset.

### 2. No GitHub Pages hosting

The script needs to be served at a stable URL such as:

```sh
curl -fsSL https://mnemosyne-systems.github.io/orangu/install.sh | sh
```

There is currently no `gh-pages` branch or Pages configuration in the repository, so this URL would 404. The hosting needs to be set up alongside the script itself.

### 3. Tarball extraction

Release assets are `.tar.gz` archives (`.zip` on Windows). The script must extract only the `orangu` binary without assuming a fixed directory structure inside the archive.

### 4. Missing tool fallbacks

The script cannot assume `curl`, `wget`, `tar`, or `jq` are all present. It needs to detect what is available and either fall back gracefully or emit a clear error naming exactly which tool is missing.

### 5. `~/.local/bin` may not exist or be in `$PATH`

On fresh installs, `~/.local/bin` often does not exist and is not in `$PATH` by default across all distros. The script must create the directory if missing and warn the user clearly if it is not on their path — with the exact line to add to their shell rc file.

### `install.sh` (Linux / macOS)

- Detect OS (`linux` / `darwin`) and architecture (`x86_64` / `aarch64`)
- On Linux, detect glibc vs musl and select the correct target triple
- Query the GitHub API for the latest release tag
- Download the matching `.tar.gz` asset
- Extract the `orangu` binary and install to `~/.local/bin` (overridable via `$INSTALL_DIR`)
- Create `~/.local/bin` if it does not exist
- Warn with the exact shell rc fix if `~/.local/bin` is not in `$PATH`
- Optionally run `orangu -s` post-install for shell completions

### `install.cmd` (Windows)

- Use `PowerShell` or `curl.exe` (available since Windows 10) to fetch the latest release
- Download `x86_64-pc-windows-msvc.zip` from the release assets
- Extract and place `orangu.exe` in `%USERPROFILE%\.local\bin` (or `%INSTALL_DIR%` if set)
- Warn if the install directory is not in `%PATH%`

### Supported targets

| OS      | Arch          | Target                       |
|---------|---------------|------------------------------|
| Linux   | x86_64 glibc  | `x86_64-unknown-linux-gnu`   |
| Linux   | x86_64 musl   | `x86_64-unknown-linux-musl`  |
| Linux   | aarch64 glibc | `aarch64-unknown-linux-gnu`  |
| Linux   | aarch64 musl  | `aarch64-unknown-linux-musl` |
| macOS   | Intel         | `x86_64-apple-darwin`        |
| macOS   | Apple Silicon | `aarch64-apple-darwin`       |
| Windows | x86_64        | `x86_64-pc-windows-msvc`     |

### Notes

- All required binary assets are already produced by release CI as of v0.6.0
- The `-s` / `--shell-completions` flag is already implemented (#98) — the installer can invoke it as an optional post-install step

tejassinghbhati:
Adds `install.sh` and `install.cmd` so users can install orangu with a single command instead of building from source.

**Linux / macOS:**
```sh
curl -fsSL https://raw.githubusercontent.com/mnemosyne-systems/orangu/main/install.sh | sh
```

**Windows:**
```cmd
curl -fsSL https://raw.githubusercontent.com/mnemosyne-systems/orangu/main/install.cmd -o install.cmd && install.cmd
```

Both scripts auto-detect the platform, fetch the latest release from the GitHub API, and install the binary to `~/.local/bin` / `%USERPROFILE%\.local\bin`. On Linux, libc variant (gnu vs musl) is detected automatically. A PATH warning is shown if the install directory is not on the user's path. `INSTALL_DIR` can be set to override the destination.

README and quickstart manual updated with install instructions.

Closes #105

### DIFF
--- a/install.cmd
+++ b/install.cmd
@@ -1,0 +1,54 @@
+@echo off
+:: Copyright (C) 2026 The orangu community
+::
+:: This program is free software: you can redistribute it and/or modify
+:: it under the terms of the GNU General Public License as published by
+:: the Free Software Foundation, either version 3 of the License, or
+:: (at your option) any later version.
+::
+:: This program is distributed in the hope that it will be useful,
+:: but WITHOUT ANY WARRANTY; without even the implied warranty of
+:: MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+:: GNU General Public License for more details.
+::
+:: You should have received a copy of the GNU General Public License
+:: along with this program. If not, see <https://www.gnu.org/licenses/>.
+:: Usage: curl -fsSL https://raw.githubusercontent.com/mnemosyne-systems/orangu/main/install.cmd -o install.cmd && install.cmd
+:: Override install directory: set "INSTALL_DIR=C:\Tools" && install.cmd
+setlocal EnableDelayedExpansion
+
+set "REPO=mnemosyne-systems/orangu"
+if not defined INSTALL_DIR set "INSTALL_DIR=%USERPROFILE%\.local\bin"
+set "TMP=%TEMP%\orangu-install-%RANDOM%%RANDOM%"
+
+where powershell >nul 2>&1 || (echo error: PowerShell is required & exit /b 1)
+
+echo Fetching latest release...
+for /f "usebackq delims=" %%v in (`powershell -NoProfile -NonInteractive -Command ^
+    "(Invoke-RestMethod 'https://api.github.com/repos/%REPO%/releases/latest').tag_name"`) do set "VERSION=%%v"
+if "!VERSION!"=="" (echo error: could not fetch latest release & exit /b 1)
+echo Version: !VERSION!
+
+set "ASSET=orangu-!VERSION!-x86_64-pc-windows-msvc.zip"
+set "URL=https://github.com/%REPO%/releases/download/!VERSION!/!ASSET!"
+
+echo Downloading !ASSET!...
+mkdir "!TMP!" >nul 2>&1
+powershell -NoProfile -NonInteractive -Command ^
+    "Invoke-WebRequest -Uri '!URL!' -OutFile '!TMP!\!ASSET!' -UseBasicParsing" || ^
+    (echo error: download failed & rd /s /q "!TMP!" >nul 2>&1 & exit /b 1)
+
+powershell -NoProfile -NonInteractive -Command ^
+    "$ErrorActionPreference='Stop'; Expand-Archive -Path '!TMP!\!ASSET!' -DestinationPath '!TMP!\out' -Force" || ^
+    (echo error: could not extract archive & rd /s /q "!TMP!" >nul 2>&1 & exit /b 1)
+if not exist "!TMP!\out\orangu.exe" (echo error: binary not found in archive & rd /s /q "!TMP!" >nul 2>&1 & exit /b 1)
+
+if not exist "!INSTALL_DIR!" mkdir "!INSTALL_DIR!"
+copy /y "!TMP!\out\orangu.exe" "!INSTALL_DIR!\orangu.exe" >nul || ^
+    (echo error: could not write to !INSTALL_DIR! & exit /b 1)
+
+rd /s /q "!TMP!" >nul 2>&1
+echo Installed: !INSTALL_DIR!\orangu.exe
+echo Run "orangu --help" to get started.
+echo Run "orangu -s" and add the output to your PowerShell $PROFILE for completions.
+endlocal
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,90 @@
+#!/bin/sh
+# Copyright (C) 2026 The orangu community
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# Usage: curl -fsSL https://raw.githubusercontent.com/mnemosyne-systems/orangu/main/install.sh | sh
+# Override install directory: INSTALL_DIR=/usr/local/bin sh install.sh
+set -e
+
+REPO="mnemosyne-systems/orangu"
+INSTALL_DIR="${INSTALL_DIR:-$HOME/.local/bin}"
+
+if command -v curl >/dev/null 2>&1; then
+    fetch()    { curl -fsSL "$1"; }
+    fetch_to() { curl -fsSL -o "$2" "$1"; }
+elif command -v wget >/dev/null 2>&1; then
+    fetch()    { wget -qO- "$1"; }
+    fetch_to() { wget -qO "$2" "$1"; }
+else
+    echo "error: curl or wget is required" >&2; exit 1
+fi
+
+command -v tar >/dev/null 2>&1 || { echo "error: tar is required" >&2; exit 1; }
+
+# Detect OS and architecture
+case "$(uname -s)" in
+    Linux)  OS="linux" ;;
+    Darwin) OS="darwin" ;;
+    *)      echo "error: unsupported OS: $(uname -s)" >&2; exit 1 ;;
+esac
+
+case "$(uname -m)" in
+    x86_64)        ARCH="x86_64" ;;
+    aarch64|arm64) ARCH="aarch64" ;;
+    *)             echo "error: unsupported architecture: $(uname -m)" >&2; exit 1 ;;
+esac
+
+# On Linux, detect glibc vs musl
+if [ "$OS" = "linux" ]; then
+    if ldd --version 2>&1 | grep -qi musl; then LIBC="musl"; else LIBC="gnu"; fi
+    TARGET="${ARCH}-unknown-linux-${LIBC}"
+else
+    TARGET="${ARCH}-apple-darwin"
+fi
+
+echo "Platform: ${TARGET}"
+
+# Resolve latest release
+VERSION=$(fetch "https://api.github.com/repos/${REPO}/releases/latest" \
+    | grep '"tag_name"' | sed 's/.*"tag_name": *"\([^"]*\)".*/\1/')
+[ -n "$VERSION" ] || { echo "error: could not fetch latest release" >&2; exit 1; }
+echo "Version:  ${VERSION}"
+
+ASSET="orangu-${VERSION}-${TARGET}.tar.gz"
+URL="https://github.com/${REPO}/releases/download/${VERSION}/${ASSET}"
+
+TMP=$(mktemp -d)
+trap 'rm -rf "$TMP"' EXIT
+
+echo "Downloading ${ASSET}..."
+fetch_to "$URL" "${TMP}/${ASSET}" || { echo "error: download failed: ${URL}" >&2; exit 1; }
+
+tar -xzf "${TMP}/${ASSET}" -C "$TMP" || { echo "error: could not extract ${ASSET}" >&2; exit 1; }
+[ -f "${TMP}/orangu" ] || { echo "error: binary not found in archive" >&2; exit 1; }
+
+mkdir -p "$INSTALL_DIR"
+[ -w "$INSTALL_DIR" ] || { echo "error: ${INSTALL_DIR} is not writable — try sudo or set INSTALL_DIR" >&2; exit 1; }
+
+cp "${TMP}/orangu" "${INSTALL_DIR}/orangu"
+chmod +x "${INSTALL_DIR}/orangu"
+echo "Installed: ${INSTALL_DIR}/orangu"
+
+case ":${PATH}:" in
+    *":${INSTALL_DIR}:"*) ;;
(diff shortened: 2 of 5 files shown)


tejassinghbhati:
image


## Security: "Running unverified scripts from GitHub is a critical security oversight"

This is the standard installation pattern used across the open-source ecosystem. Projects such as Rust (`rustup`), Homebrew, Node Version Manager (`nvm`), Bun, and many others distribute installers via a `curl | sh` workflow.

The security model relies on:

- HTTPS/TLS to ensure transport integrity.
- GitHub's infrastructure and certificates to authenticate the source repository.
- User trust in the project's repository and maintainers.

Adding a checksum or signature to a one-line installer does not meaningfully improve security unless users already possess a trusted copy of the verification key. At that point, the simplicity and purpose of the bootstrap installer are lost.

This is therefore not a flaw introduced by this PR but an established industry-standard approach for bootstrap installers.

---

## Test Suite: "No test suite provided for install.sh"

This statement is factually incorrect.

`tests/install_test.sh` was added as part of this PR and was explicitly approved elsewhere in the review. The review cannot simultaneously approve the test suite and claim that no tests were provided.

---

## Test Suite: "Missing shebang line check"

This statement is also incorrect.

The shebang:

```sh
#!/bin/sh
```

is present on line 1 of `install.sh`.

The review appears to have overlooked the existing implementation.

---

## Documentation: "install.cmd assumes PowerShell is universally available without documentation"

This requirement is already documented in the script header.

Additionally, PowerShell has been included as a standard Windows component since Windows 7 and is available by default on all supported Windows versions.

The script header explicitly states:

```text
Requirements: PowerShell (included with Windows 10 and later)
```

Therefore, both the dependency and its availability are already documented.

---

## Documentation: "No explanation for why INSTALL_DIR defaults to %USERPROFILE%\.local\bin"

The chosen default intentionally mirrors the Unix convention of:

```text
~/.local/bin
```

Using:

```text
%USERPROFILE%\.local\bin
```

on Windows provides:

- A user-local installation location.
- No requirement for administrator privileges.
- Consistency with the Unix installer.
- Isolation from system-wide directories.

This is the standard and safest default for a user-facing installer.

---

## Summary

Several review comments appear to be based on incorrect assumptions or information already present in the patch:

- A test suite was provided and approved.
- The shebang already exists.
- PowerShell requirements are already documented.
- The installation directory follows established cross-platform conventions.
- The security concern describes the standard bootstrap installer model used by major open-source projects rather than a flaw unique to this PR.

For these reasons, the cited objections do not appear sufficient grounds for rejecting the patch.

tests/install_test.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
jesperpedersen, line 1: License header

install.cmd
```diff
@@ -0,0 +1,39 @@
+@echo off
```
jesperpedersen, line 1: License header

install.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
```
jesperpedersen, line 1: License header

jesperpedersen:
## Overall

- The change set implements a unified installation workflow for Windows (install.cmd) and Unix (install.sh) with consistent logic for fetching releases and managing paths, despite install.cmd being rejected due to critical security and concurrency flaws.
- Common themes include fragile external dependency parsing (using grep/sed for JSON and %RANDOM% for temp dirs) and a lack of checksum verification, exposing users to MITM and collision risks.
- Risk is elevated by missing error handling in both scripts for network failures, extraction errors, and permission issues, leading to confusing downstream failures rather than clear user feedback.
- Test coverage is insufficient, focusing on trivial syntax checks and hardcoded strings rather than validating the actual installation logic, error paths, or environment variable handling.
- Documentation contains syntax errors in the Windows one-liner and ambiguous PowerShell completion instructions, requiring correction to prevent user installation failures.

## Code

- **install.cmd:26**: Race condition and logic error: `set "TMP=%TEMP%\orangu-install-%RANDOM%"` allows collisions if multiple instances run simultaneously; use `mklink` or `powershell -command [System.IO.Path]::GetRandomFileName()` for uniqueness.
- **install.cmd:26**: Security/Correctness: Using `%RANDOM%` (0-32767) creates predictable and non-unique temporary directory names, leading to potential file overwrites or collisions in concurrent executions.
- **install.cmd:30**: Error handling: The `for` loop fetching the version does not handle HTTP errors (e.g., 403 rate limit, 404 private repo) gracefully; `Invoke-RestMethod` might throw an exception that isn't caught by the `||` operator in this specific context, or might return an empty string which is caught, but the error message is generic.
- **install.cmd:38**: Security/Correctness: Passing `!URL!` and `!ASSET!` directly into PowerShell command strings without proper escaping could lead to command injection if the version string contains special characters (e.g., quotes, semicolons).
- **install.cmd:42**: Error handling: `Expand-Archive` failure is not checked with an `if errorlevel` check; if the archive is corrupt or invalid, the script continues and fails later with a confusing error message.
- **install.sh:56-57**: Parsing GitHub API JSON with grep/sed is fragile and may break if the API response format changes or includes unexpected whitespace; consider using `jq` if available or a more robust parser.
- **install.sh:64**: `VERSION` is not quoted in the assignment context, though it is safe here due to the check on line 65; quoting is generally preferred for robustness against whitespace in variable values.
- **install.sh:71**: `tar` extraction is not checked for errors (missing `||` clause); if extraction fails, the script continues and may produce misleading errors later.
- **install.sh:82**: PATH check logic is correct but could be simplified; however, the current implementation is functionally sound.
- **install.sh:85-86**: No error handling if `orangu --help` or `orangu -s` commands are not found after installation; this is a minor usability issue rather than a code defect.

## Security

- **install.cmd:29**: Potential command injection via !VERSION! if the GitHub API returns a malicious tag name; however, the tag is used in a file path/URL context which is largely sanitized by URL encoding, but still a minor risk if the server blindly concatenates.
- **install.cmd:34**: The URL constructed from the GitHub API response is used directly in Invoke-WebRequest; while Invoke-WebRequest handles most injection, relying on unvalidated external input for the base URL structure is a slight anti-pattern, though low risk here.
- **install.cmd:38**: The downloaded archive is extracted to a temporary directory before copying, which is a good security practice (mitigates direct execution of malicious binaries).
- **install.cmd:41**: The installation directory is created if it doesn't exist, which is standard behavior.
- **install.cmd:44**: Cleanup of temporary files is performed, reducing risk of leftover malicious artifacts.
- **install.sh:66-67**: Fetching GitHub API response and parsing JSON using grep/sed is fragile and prone to injection if tag names contain special characters; however, the tag is used in a controlled URL path and not executed as shell code, making the risk low but non-zero.
- **install.sh:66-67**: No JSON parsing library is used; if the API response format changes or contains unexpected whitespace/escaping, the script may fail or extract incorrect version strings, leading to installation of wrong binaries.
- **install.sh:80-81**: Downloading and executing (via tar extraction) binaries from the internet without checksum verification (e.g., GPG signature or SHA256) exposes the user to MITM attacks if the GitHub CDN or network is compromised.
- **install.sh:33-35**: The script does not pin the GitHub API endpoint to a specific TLS version or certificate, relying on system defaults which may be outdated or vulnerable to downgrade attacks on older systems.
- **install.sh:76-77**: The temporary directory is created with default permissions; while `mktemp -d` is generally secure, the subsequent extraction and copy steps do not enforce strict permissions on the binary before installation, though `chmod +x` is applied at the end.

## Memory

No issues found

## Performance

- **install.cmd:35-36**: Inefficiently calls PowerShell twice to fetch the version (once for the command, once for the variable assignment logic could be streamlined, but acceptable).
- **install.cmd:40-42**: Inefficiently calls PowerShell twice for download and extraction; combining into a single PowerShell script block would reduce process startup overhead.
- **install.cmd:48**: `rd /s /q` is called unconditionally; while harmless, it adds minor I/O overhead if the directory was already removed or failed to create.
- **install.cmd:49-50**: Echoing help text is not a performance issue, but checking for existing installation before copy (line 46-48) could save I/O if the file hasn't changed.
- **install.sh:63**: Minor performance inefficiency; parsing JSON via `grep`/`sed` is fragile but acceptable for a single-field extraction, though `jq` would be more robust and potentially faster for complex structures. No critical performance issue.
- **install.sh:68**: No significant performance issue.
- **install.sh:73**: No significant performance issue.
- **install.sh:77**: No significant performance issue.
- **install.sh:81**: No significant performance issue.

## Test Suite

- **install.cmd:32-34**: Missing test coverage for error handling when PowerShell is unavailable or fails to execute.
- **install.cmd:35-38**: Missing test coverage for failure cases when fetching the latest release version (e.g., network errors, empty response).
- **install.cmd:44-46**: Missing test coverage for download failures, including network timeouts and HTTP errors.
- **install.cmd:48-50**: Missing test coverage for archive extraction failures or missing binary in the downloaded zip.
- **install.cmd:52-54**: Missing test coverage for permission errors when writing to the installation directory.
- **tests/install_test.sh:33-35**: Hardcoded expected strings for target triples bypass the logic under test; these tests verify static strings rather than the `install.sh` logic for constructing triples.
- **tests/install_test.sh:38-41**: The `INSTALL_DIR` override test is redundant with the default value logic tested on line 38, as `${INSTALL_DIR:-default}` is a standard shell feature not specific to the script's implementation.
- **tests/install_test.sh:44-45**: The "temp dir cleaned up" test is logically trivial and checks shell syntax (`[ ! -d ]`) rather than script behavior, as `rm -rf` guarantees non-existence.
- **tests/install_test.sh:49**: The final exit check `[ "$fail" -eq 0 ]` relies on `set -e` (line 18) to exit non-zero on failure, but if `set -e` were removed, this line would need to explicitly `exit 1` to signal failure to the caller.

## Documentation

- **README.md:65-68**: Documentation error: The command `set INSTALL_DIR=C:\Tools && install.cmd` will fail because the variable is not exported to the script's environment; it requires `set INSTALL_DIR=C:\Tools` on one line and `install.cmd` on the next, or using `cmd /c "set ... & install.cmd"`.
- **README.md:74**: Ambiguous instruction: "add `Invoke-Expression (orangu -s)`" is syntactically incorrect for PowerShell (should be `Invoke-Expression (orangu -s)` or just `orangu -s | Invoke-Expression` depending on intent, but usually `orangu -s` outputs the script content which needs to be invoked). It should likely be `Invoke-Expression (orangu -s)` which is valid, but the phrasing is slightly confusing compared to the sh examples. More critically, `orangu -s` likely outputs the completion script source code, so `Invoke-Expression` is correct, but the user might copy-paste the output directly. The example `orangu -s >> ~/.bashrc` implies `orangu -s` outputs the script. The PowerShell example `Invoke-Expression (orangu -s)` is correct syntax. However, line 65 is a definite error.
- **README.md:65**: Security/Usability: The one-liner install script uses `curl` to pipe to `sh`. While common, it lacks a checksum verification step in the documentation, which is a best practice for install scripts. This is a minor documentation gap regarding security.
- **README.md:70**: Inconsistency: The shell completion instructions for bash and zsh append to the config file (`>>`) and then source it. For fish, it pipes to `source`. This is technically correct for each shell, but the visual inconsistency might confuse users. Not a critical error.
- **README.md:53**: Missing context: The section "One-liner install" mentions `install.sh` and `install.cmd` but does not provide a link to their source code or a disclaimer about their contents, which is a minor documentation transparency issue.
- **README.md**: Primary finding is line 65.
- **README.md:65**: Syntax error: `set INSTALL_DIR=C:\Tools && install.cmd` does not export the variable to the child process in standard CMD; it requires two separate commands or a subshell.
- **install.cmd:24**: The usage comment on line 24 suggests using `curl` to download the script, but does not mention that the script itself requires PowerShell to be present and accessible in the PATH.
- **install.cmd:24**: The usage comment does not specify the required TLS version or certificate validation behavior, which might be relevant for users in restricted corporate environments (though `Invoke-WebRequest` handles this reasonably by default).
- **install.cmd:32**: The error message "could not fetch latest release" is generic; it does not distinguish between network errors, API rate limiting, or invalid JSON responses, which might help users troubleshoot.
- **install.cmd:40**: The error message "download failed" is generic and does not indicate if the failure was due to network issues, 404 (asset missing), or other HTTP errors.
- **install.cmd:48**: The error message "could not write to !INSTALL_DIR!" does not explicitly suggest running as Administrator or checking permissions, which is the most common cause of failure for this step.

## Conclusion

**orangu rejects this patch**

- Rejected: **README.md**
- Rejected: **install.cmd**

Generated by: **orangu 0.7.0** (bartowski/Qwen_Qwen3.6-35B-A3B-GGUF)

jesperpedersen:
@tejassinghbhati See how many you want to fix - PTAL me when you are happy so we can get 0.7 out

tejassinghbhati:
**Fixed:**
- `install.cmd` — extraction now guarded (`-ErrorAction Stop` + cleanup on failure), temp dir gets more entropy
- `install.sh` — explicit error on `tar` extraction failure
- README + `install.cmd` — fixed the Windows override one-liner: `set "INSTALL_DIR=C:\Tools" && install.cmd`
- `install_test.sh` — triple tests now actually build the triple instead of comparing identical strings

**Skipped (out of scope for a minimal installer):**
- checksums/GPG — nothing to verify against until releases publish them (good follow-up)
- TLS pinning — worse than the system default
- jq — keeping it zero-dependency like rustup/nvm

ptal, i think this is okay now

jesperpedersen:
Rebased and merged.

PR #108, opened 2026-06-17
[#30] Workspace model and manager

tejassinghbhati:
Add the workspace tab model that lets several projects stay open in one orangu instance instead of running one instance per project.

WorkspaceManager owns the ordered list of open workspaces and the active tab, with open/close/switch/next/prev. It keeps two invariants from the design: there is always at least one workspace (closing the last is rejected, only /quit ends orangu), and closing a tab renumbers the ones after it. Focus follows sensibly when the active tab is closed.

This is the model only; the /workspace command, key bindings, the status line and per-tab sessions build on top of it in later steps.


jesperpedersen:
@tejassinghbhati Looks like a good start.

We should likely have 

```
workspaces=top|bottom|left|right
```

in `orangu.conf` to place the workspace tabs where the user wants them

jesperpedersen:
@tejassinghbhati Remember `orangu -i` with completion support

jesperpedersen:
@tejassinghbhati Lets continue with the Alt shortcuts and documentation

tejassinghbhati:
Starting on the Alt shortcuts.

One thing first: the keys need the in-process tab runtime underneath them, several workspaces live in one orangu, each its own session and scrollback so it's a fair bit more than just key handling, more a refactor of the main loop.

Since #108 is in good shape on its own (the model, config, the wizard and the /workspace command), I'd rather land the tab runtime + Alt keys + status line as a separate PR on top of it, instead of growing #108 while you're reviewing it.
Keeps #108 mergeable now and gives the bigger change its own review. The docs would go with that PR too.

Sound good? I'll keep the 🟢/🔴 validity and per-workspace model for a later one, like we said.

jesperpedersen:
@tejassinghbhati Ok, add the documentation for #108 first. You can take 31-workspaces.md for it - I'll do 32-review.md

jesperpedersen:
Merged.

PR #109, opened 2026-06-17
Issue 30 workspace command

tejassinghbhati:
Part of #30, following the split started in #108.

Adds the `/workspace` command for opening and switching workspaces, both as a
slash command and in natural language (`workspace`, `switch workspace`,
`workspace 1`).

A bare number is a tab; anything else is a directory:

- `/workspace` — report the active workspace
- `/workspace 1` — switch to a tab by its number
- `/workspace ~/project` — open a directory, or report it as already active

`/workspace <path>` re-execs into the workspace the same way `/session <path>`
already opens one, so it auto-resumes a session there if one exists. Tab
completion offers the workspaces seen in past sessions first, then falls back to
filesystem directory completion.

Until tabs share a single process, the active workspace is the only one open, so
it is tab 1 and any higher number has no tab to switch to yet. The live tab list
it switches among lands with the in-process tabs.

Help text and parse/dispatch/completion tests are included.

---

Stacked on #108 (the workspace model). The diff here includes that commit until
#108 merges, so it's best reviewed after, or alongside it.


jesperpedersen:
@tejassinghbhati Please, merge this into #108

tejassinghbhati:
Merged into #108 
therefore closing
thankyou!

PR #114, opened 2026-06-19
[#30] Workspace tabs: live switching, Alt keys, status line

tejassinghbhati:
Make workspace tabs live in one orangu instead of re-execing per switch. Several workspaces stay open at once, each its own session, conversation, scrollback, pending queue and history.

The run loop keeps the active tab's state in its existing locals; a WorkspaceRing parks the other tabs and swaps the active one in and out, so the loop body is unchanged. Switching is deferred to the top of the next iteration so it never races the render context's borrows.

Keys at the prompt: Alt+, / Alt+. move to the previous/next tab (wrapping), Alt+Insert starts a new workspace (pre-filling /workspace), Alt+Delete closes the active tab (the last tab is never closed; only /quit exits). /workspace now switches and opens live tabs instead of re-execing: a number selects a tab, a directory opens one (or switches if already open).

A tab bar under the banner shows the open tabs with the active one bold, once more than one is open. On exit every open tab is saved and prints its own resume line annotated with its workspace and branch.

WorkspaceManager is made generic over the tab payload (WorkspacePath) so the binary's rich tab type reuses the tested bookkeeping.

The status-bar tab placement (workspaces=bottom|left|right) and per-tab server/model are follow-ups.

### DIFF
--- a/doc/manual/en/03-quickstart.md
+++ b/doc/manual/en/03-quickstart.md
@@ -173,7 +173,7 @@ comment on 42 with auto review  # post it on GitHub/GitLab issue/PR #42
 
 By default the tools operate on the current directory. Use `--workspace /path/to/project` to point **orangu** at another tree.
 
-The startup flags have short forms: `-c` for `--config`, `-w` for `--workspace`, `-r` for `--resume`, and `-i` for `--init`.
+The startup flags have short forms: `-c` for `--config`, `-w` for `--workspace`, `-r` for `--resume`, `-i` for `--init`, and `-a` for `--all` (reopen the tabs from the previous run).
 
 **orangu** automatically resumes an existing session when you return to the same workspace and Git branch. When a previous session is found, the status bar shows:
 
--- a/doc/manual/en/31-workspaces.md
+++ b/doc/manual/en/31-workspaces.md
@@ -4,9 +4,11 @@
 
 A *workspace* is the project directory orangu is open on. It is the directory the model's tools read from and write to, the directory whose Git branch drives the prompt and the Git and review commands, and the directory a session belongs to. By default the workspace is the directory you launched orangu in; pass `--workspace <path>` (or `-w <path>`) at startup to begin somewhere else.
 
-The `/workspace` command reports the active workspace and switches between workspaces without leaving the client. Each workspace is its own session, so switching to a workspace resumes the session you last had open there, or starts a fresh one when there is none. See the Sessions section of the Terminal interface chapter for how sessions are matched to a workspace and branch.
+Several workspaces can be open at once as tabs in a single orangu, instead of running one instance per project. Each tab is its own session — its own conversation, scrollback, pending queue and command history — so switching tabs returns to exactly where you left that workspace, or resumes the session it last had. A tab bar shows the open tabs with the active one in bold, placed where the `workspaces` setting puts it (see Tab placement); it appears once more than one workspace is open. When `feedback` is on, a colored dot precedes each tab number: green for a valid tab, red when the tab's branch has been deleted or changed underneath it, and a blinking white dot on the active tab while it is processing a response. See the Sessions section of the Terminal interface chapter for how sessions are matched to a workspace and branch.
 
-Where the workspace tabs are drawn is set by the `workspaces` key in the `[orangu]` configuration section; see the Tab placement section below.
+Pass `-a` or `--all` at startup to reopen the tabs that were open at the end of the previous run. Their workspace directories and auto-resumed sessions are restored behind the initial tab.
+
+You move between tabs with the `/workspace` command or the workspace key bindings, and choose where the tab bar sits with the `workspaces` configuration key — both are described below.
 
 \newpage
 
@@ -14,7 +16,7 @@ Where the workspace tabs are drawn is set by the `workspaces` key in the `[orang
 
 With no argument, `/workspace` reports the active workspace directory.
 
-A bare number selects a workspace by its tab number; anything else is treated as a directory path. Switching to a directory opens it as a workspace and resumes the matching session there, or starts a fresh one if there is none.
+A bare number switches to the open tab with that number. Anything else is treated as a directory path: orangu switches to the tab already open on it, or opens a new tab for it (resuming the matching session there, or starting a fresh one). `Alt+Insert` opens a fresh tab on the current directory immediately; you then re-point it with `/workspace <path>`.
 
 Tab completion offers the workspaces seen in earlier sessions first, then completes filesystem directories, so a directory that has never been opened can still be navigated to a segment at a time:
 
@@ -43,13 +45,35 @@ switch workspace ~/projects/orangu
 
 \newpage
 
+## Key bindings
+
+At the prompt, the workspace tabs are driven by these keys:
+
+| Key | Action |
+| :-- | :-- |
+| `Alt+,` | Switch to the previous tab (to the left), wrapping |
+| `Alt+.` | Switch to the next tab (to the right), wrapping |
+| `Alt+Insert` | Open a new workspace tab (re-point it with `/workspace`) |
+| `Alt+Delete` | Close the active tab |
+
+Closing a tab renumbers the ones after it and moves focus to a neighbour. The last open tab is never closed — only `/quit` ends orangu. The keys work at the prompt and during streaming: pressing one while the model is responding parks the stream in the background and switches immediately. The response keeps running in the original tab; a blinking dot in the tab bar shows it is still active. Switching back re-attaches the live view so you can watch it complete. Press Escape twice to cancel a response whether it is running in the foreground or in the background.
+
+\newpage
+
 ## Tab placement
 
-Where the workspace tabs are shown is set by the `workspaces` key in the `[orangu]` section of the configuration file:
+Where the workspace tab bar is shown is set by the `workspaces` key in the `[orangu]` section of the configuration file:
 
 ```ini
 [orangu]
 workspaces = top
 ```
 
-The options are `top` (the default), `bottom`, `left`, and `right`, matched case-insensitively. The interactive `--init` wizard prompts for it with Tab completion. Leaving the key out uses `top`; an unrecognised value is rejected when the configuration is loaded. The full list of `[orangu]` keys is in the Configuration chapter.
+The options, matched case-insensitively, are:
+
+- `top` (the default) — a horizontal bar across the top of the screen, above the banner.
+- `bottom` — a horizontal bar across the bottom, under the input.
+- `left` — a vertical column of tab numbers down the left edge, beside the banner and output.
+- `right` — the same column down the right edge.
+
+The interactive `--init` wizard prompts for it with Tab completion. Leaving the key out uses `top`; an unrecognised value is rejected when the configuration is loaded. The full list of `[orangu]` keys is in the Configuration chapter.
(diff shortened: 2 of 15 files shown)


jesperpedersen:
```
workspaces = X
```

doesn't appear to be working.

And, the workspace tabs should be against the terminal. So

Top
```
1 | 2
-------
<BANNER>
```

Bottom
```
-------
<INPUT>
-------
1 | 2
```

Left
```
1 | <BANNER>
2 |
```

Right
```
<BANNER> | 1
         | 2
```

jesperpedersen:
For `Alt+<INSERT>` can just start the new workspace tab then the user can switch server and workspace after

jesperpedersen:
If feedback is on then we need a green, red or white (blinking) dot for each workspace tab

jesperpedersen:
We need an `orangu -a|--all` option to start with the workspaces active in the last run of orangu - save that in ~/.orangu/workspaces. So, that need to be updated upon `/quit`

jesperpedersen:
@tejassinghbhati Are all my comments in this commit ?

tejassinghbhati:
Not yet, only the placement one (top/bottom/left/right) is in this commit.

Still to come from your other comments:

- Alt+Insert just starting a new tab instead of pre-filling /workspace. The
  "switch server after" part needs per-workspace /server and /model, which is
  the follow-up we said we'd leave till the end, so I'll land the direct
  new-tab now and the per-tab server/model with that.
- The green / red / white-blinking dot per tab when feedback is on.
- orangu -a|--all to reopen the last run's workspaces, saved in
  ~/.orangu/workspaces and updated on /quit.

I'll add those next. Wanted to keep the placement fix on its own first.

tejassinghbhati (commented):
Alt+Insert now opens a blank fresh tab with its own new session instead of pre-filling /workspace in the input. You re-point it with /workspace <path> once it opens.

-a / --all flag: on quit, the open workspace paths are written to ~/.orangu/workspaces. Starting with orangu -a reads that file back and reopens each tab behind the initial one, auto-resuming their sessions.

jesperpedersen:
Can you use the "ghost" color for the non-selected workspace ? Also, try "Hello, who are you ?" - then you can't use the Alt commands for workspace

In bottom mode the cursor isn't placed inside the input area.

jesperpedersen:
@tejassinghbhati Test `/review` and `/auto_review` mode as well

tejassinghbhati (commented):
- Ghost color: non-active tab numbers now render in dim gray in both the horizontal bar and vertical gutter, so the active tab stands out with bold.
- Alt keys during streaming: instead of dropping the keypress, the switch is now queued and applied as soon as the response finishes (doc updated to match).
- Bottom mode cursor: after writing the bar at the last row, the cursor is repositioned back to the input area.

jesperpedersen:
@tejassinghbhati We need to find a way to switch between workspaces while streaming - we can't wait for a response because that can take hours... I think we need to hook "into" the status line, that updates each second

jesperpedersen:
@tejassinghbhati Or the status area for `/auto_review`

jesperpedersen:
@tejassinghbhati Something is wrong

```
Who are you ?
<THINKING>
Alt+<INSERT>
Alt+,
Request cancelled
````

**Comment by tejassinghbhati** [comment]
You're right, my approach was wrong, cancelling the stream on a tab switch is not the right behavior.
The correct fix perhaps is to let the LLM keep streaming in the background while you switch tabs. 

That requires running the stream as a background task (tokio spawn) so wait_for_response can park the active tab into the ring, unpack the target tab, and continue rendering from it, while the stream writes back to the original tab's output state asynchronously.

That's a bigger architectural change. I'll look into it.

**Comment by jesperpedersen** [comment]
@tejassinghbhati You have design it as completely different environments that operate independently.

```
WS1:
Who are you ?
Alt+INSERT

WS2:
/server code-main
Implement a B+ tree in Rust
Alt+INSERT
```

and so on...  All workspaces work in parallel

**Comment by jesperpedersen** [comment]
@tejassinghbhati PTAL me once you are happy with it

**Comment by jesperpedersen** [comment]
@tejassinghbhati Make sure you are rebased - I have pushed some big patches

**Review by tejassinghbhati — COMMENTED** [review]
Final Review

### Status
Build, fmt, 232 tests pass - done

### What ships
- Workspace tabs: each tab is an independent session (conversation, tools, skills, branch)
- True background streaming: tab-switch mid-LLM parks the task in `tokio::spawn`, keeps running
- Re-attaching renders the live stream from where it is; ESC×2 cancels fore or background
- `Alt+,` / `Alt+.` / `Alt+Insert` / `Alt+Delete` key bindings
- `/workspace <path|n>`, `-a`/`--all` startup flag, tab bar (top/bottom/left/right)

### What was fixed since first push
- Resolved two rounds of upstream merge conflicts (#87 skills, #119 auto_review link, #120 prestart, #121 branch indent)
- `deferred_tab` type narrowed from `Option<InputResult>` → `Option<TabAction>` (removed `input_result_to_tab_action`)
- Skills scoped per-tab; `WorkspaceTab::open()` discovers and embeds them into the session prompt
- 6 unit tests added for `WorkspaceRing` (rotate wrap, close renumber, last-tab guard, `position_of`)
- Doc gaps patched: `-a`/`--all` added to startup flags table and quickstart short-flags sentence

### Remaining known item
`WorkspaceTab` has 18 `pub(crate)` fields: acceptable for now, natural refactor candidate later.

### Verdict
Ready to merge.

**Comment by jesperpedersen** [comment]
@tejassinghbhati And, now another rebase

**Comment by jesperpedersen** [comment]
## Overall

- Implements a new workspace tab management system including persistence, navigation macros, and TUI updates.
- High readiness for core logic and documentation, but requires cleanup of redundant code and off-by-one UI calculations.
- Moderate risk due to potential deadlocks in `wait.rs` and unhandled filesystem errors in `session_store.rs`.
- Common themes include manual state management via macros, heavy string allocations in the TUI layer, and a lack of unit tests for new traits.

## Code

- **src/bin/orangu/input.rs:119-122**: `tab_statuses` is passed as a slice, but its relationship to `tab_bar`'s internal state is not explicitly enforced by the type system.
- **src/bin/orangu/input.rs:282-285**: The `Alt+Delete` check excludes `Control` but does not explicitly check for `Shift`, which might lead to inconsistent behavior compared to `Alt+,`.
- **src/bin/orangu/main.rs:192-202**: The `current_tab!` macro captures variables by value/move from the outer scope, which is correct for the logic but creates a large closure-like structure in the macro.
- **src/bin/orangu/main.rs:329-335**: The `load_tab!` macro uses assignment to variables that are technically moved out of the `WorkspaceTab` struct, which is the intended pattern for this "manual" state management.
- **src/bin/orangu/main.rs:522-535**: The `save_all_tabs!` macro correctly handles the persistence of both the active tab and the parked ring.
- **src/bin/orangu/main.rs:770-782**: The `InputResult` variants for workspace navigation correctly map to the `pending_tab_action` state.
- **src/bin/orangu/main.rs:882-891**: The `CommandOutcome` variants for workspace switching correctly update the `pending_tab_action` state.
- **src/bin/orangu/session_store.rs:45-47**: `std::fs::create_dir_all` is called without checking for errors, though this is acceptable given the comment that failing to save must not block exit.
- **src/bin/orangu/session_store.rs:52-56**: Using `workspace.display().to_string()` is correct for handling non-UTF8 paths, but `join("\n")` might create a trailing newline or inconsistent formatting compared to `writeln!`.
- **src/bin/orangu/session_store.rs:74-75**: `workspace.is_dir()` performs a syscall; if the file system is slow or the path is a network mount, this could slightly delay startup during `load_open_workspaces`.
- **src/bin/orangu/wait.rs:44-56**: Potential for deadlock or high contention if the `Mutex` is held while `prompt` performs long-running operations, though the closure suggests it's only held during the push/merge/update.
- **src/bin/orangu/wait.rs:177-180**: `*session = real_session_back;` correctly restores the session state from the background task.
- **src/bin/orangu/wait.rs:229-245**: The use of `handle.abort()` and `restored.restore(saved_messages)` correctly handles the state cleanup for a cancelled task.
- **src/bin/orangu/wait.rs:297-313**: Correctly handles workspace navigation by capturing the current `PendingResponse` state.
- **src/bin/orangu/wait.rs:394-409**: Correctly handles workspace navigation in the streaming command loop.
- **src/bin/orangu/workspace_tab.rs:167-170**: The `workspace_branch_name` function is used but its source is not visible in this file; ensure it is correctly imported from `crate::*`.
- **src/bin/orangu/workspace_tab.rs:251-253**: `drop(active)` is redundant as `active` is not used after this point and will be dropped at the end of the scope.
- **src/bin/orangu/workspace_tab.rs:366-374**: The `workspaces` helper function contains a complex `fold` and `map` chain that is immediately superseded by a simpler implementation; it should be cleaned up or removed.
- **src/tui/screen.rs:187**: `horizontal_tab_bar` is called with `actual_width` but uses `clip_line` internally which might not account for the gutter width in some contexts.
- **src/tui/screen.rs:282**: The calculation for `pf_top_row` and `pf_input_start` is complex and prone to off-by-one errors when calculating the cursor position for the bottom bar.
- **src/tui/screen.rs:282-288**: Logic assumes `frame_height` and `banner_rows` are sufficient to calculate the input area, but if `banner_rows` or `input_lines` change, the hardcoded `+ 1` or `+ 2` might misalign the cursor.

## Security

No issues found

## Memory

No issues found

## Performance

- **src/bin/orangu/session_store.rs:55-59**: `collect::<Vec<_>>().join("\n")` creates an intermediate allocation; can be written directly to the file using a buffered writer or a single string builder.
- **src/tui/screen.rs:126-130**: `upper.join("\r\n")` creates a new string allocation for the entire output buffer.
- **src/tui/screen.rs:162-170**: `screen.push_str(&" ".repeat(pad))` creates a new string allocation for every row in the gutter.
- **src/tui/screen.rs:190-193**: `horizontal_tab_bar` is called inside `render_screen` and then potentially again if logic branches, but it performs `Vec` collection and `join` on every call.

## Test Suite

- **src/bin/orangu/session_store.rs:44-53**: No unit tests provided to verify directory creation, file writing, or path handling.
- **src/bin/orangu/session_store.rs:58-71**: No unit tests provided to verify file reading, line parsing, and directory existence filtering.
- **src/workspaces.rs:106-112**: Missing unit tests for the new `WorkspacePath` trait and its implementation on `Workspace`.
- **src/workspaces.rs:149-153**: Missing unit tests for the generic `WorkspaceManager<W>` methods (`new`, `active`, `active_mut`, `workspaces`, `get`).
- **src/workspaces.rs:175-188**: Missing unit tests for the new `WorkspaceManager<W: WorkspacePath>` methods (`position_of`, `open_or_switch`).

## Documentation

- **src/bin/orangu/main.rs:120-123**: Documentation for `--all` is clear and fits the context of the new workspace tab logic.
- **src/bin/orangu/main.rs:345-351**: Macro documentation correctly explains the packing/unpacking mechanism for the new tab state management.
- **src/bin/orangu/main.rs:538-542**: Documentation for `parked_tab_status` accurately describes the logic for determining tab status during feedback.
- **src/bin/orangu/test_support.rs:63**: Missing documentation for the new `tab_bar` field.
- **src/bin/orangu/test_support.rs:64**: Missing documentation for the new `tab_statuses` field.

## Conclusion

- Rejected: **src/bin/orangu/session_store.rs**
- Rejected: **src/bin/orangu/test_support.rs**
- Rejected: **src/tui/screen.rs**
- Rejected: **src/workspaces.rs**

Generated by: **[orangu](https://mnemosyne-systems.github.io/orangu/) 0.8.0** (bartowski/gemma-4-12B-it-GGUF)

**Comment by jesperpedersen** [comment]
Merged.

**Comment by jesperpedersen** [comment]
@tejassinghbhati Hmm, there seems to be something wrong it the workspace colors... Do

```
WS1:
Hello ?
Alt+INSERT

WS2:
Hello ?
```

After they are done one workspace switches to :red_circle: and they alternate based on your view... They should both be :green_circle: - :red_circle: mean no access to llama.cpp, :green_circle: mean access to llama.cpp, blinking :white_circle: means working

tejassinghbhati:
Yeah I actually found that issue, please let me explain and get it solved in the very next pr, would take just a moment. Apologies for the inconvenience.

Issue #87, opened 2026-06-07
Agent Skills


jesperpedersen:
https://agentskills.io/home

sksingh2005:
@jesperpedersen anyone working on this right now ?

jesperpedersen:
@sksingh2005 Go for it !

jesperpedersen:
@sksingh2005 Probably best to start with a high-level design

sksingh2005:
## Core Concepts
* **Skill**: A custom command that would be defined by a `SKILL.md` file. It will contain a `name`, a `description` (parsed from YAML frontmatter), and a markdown body.
* **SkillRegistry**: We propose a central repository designed to discover, load, and deduplicate skills at startup.

## Discovery & Registration
We plan to load skills upon application startup by scanning specific directories:
* **User scope**: `~/.orangu/skills/*/SKILL.md`
* **Project scope**: `$WORKSPACE/.orangu/skills/*/SKILL.md`

## Architectural Components
The proposed implementation would touch the following areas in the existing architecture:

### A. Core Module (`src/skills.rs`)
We will introduce a new core module to handle the `Skill` struct and `SkillRegistry`. It would utilize `walkdir` to traverse the directories and `serde_yaml` to parse frontmatter. It is planned to expose:
* `SkillRegistry::discover(workspace_path)`
* `SkillRegistry::find(name)`
* `SkillRegistry::all()`
* `SkillRegistry::system_prompt_index()`: To render the LLM-facing list of available skills.

### B. LLM Context Injection 
To ensure the LLM is aware of the loaded skills so it can autonomously invoke them, we propose:
* Initializing the `SkillRegistry` once in `main.rs`.
* Creating a wrapper `system_prompt_with_skills` to append the rendered `system_prompt_index()` from the registry to the agent's base system prompt.

### C. Execution Flow 
When the user or the LLM types a command (e.g. `/code-review security vulns`), the `parse_local_command` fallback logic would inspect the input:
1. If the input starts with `/` and doesn't match a built-in `LocalCommand`.
2. It will look up the parsed command name in `context.skills`.
3. If a skill is found, it will render the body (replacing `$ARGUMENTS` with the provided arguments).
4. The rendered text will then be directly injected into the chat context via `session.push_user_message()`.

### D. User Interface & Autocomplete
* **`/skills` Command**: We will add a new built-in command (`LocalCommand::Skills`) that will list all active skills, their descriptions, and their source (user vs. project).
[...]

PR #122, opened 2026-06-20
[#87] Document writing agent skills with helper code

sksingh2005:


doc/manual/en/32-skills.md
```diff
@@ -0,0 +1,136 @@
...
+changes. If anything is missing, report it before suggesting a release command.
+```
+
+## Skill with helper files
+
+A skill directory may contain extra files such as examples, templates, scripts,
+or source code:
+
+```text
+.agents/skills/schema-change/
```
jesperpedersen: schema-change ?

doc/manual/en/32-skills.md
```diff
@@ -0,0 +1,136 @@
...
+name: log-parser
+description: Parse service logs with the bundled helper
+---
+
+# Log Parser
+
+When log parsing is needed, build the helper first:
+
+```sh
+rustc .agents/skills/log-parser/helper.rs -o /tmp/orangu-log-parser
```
jesperpedersen: Why /tmp ?

doc/manual/en/32-skills.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+Then run it against the user-provided log path:
+
+```sh
+/tmp/orangu-log-parser <path>
```
jesperpedersen: How does SKILL.md get hold of the binary ?

doc/manual/en/32-skills.md
sksingh2005: I think release check would be better ?

doc/manual/en/32-skills.md
sksingh2005: Sorry the example was unclear because it built the helper into `/tmp` and then referenced that path separately. I will update it to build the binary inside the skill directory and run it from there:
`.agents/skills/log-parser/log-parser
`
 That will make the path explicit in `SKILL.md` and avoids relying on an external temporary location.

doc/manual/en/32-skills.md
jesperpedersen: @sksingh2005 Yeah

jesperpedersen:
Merged.

PR #125, opened 2026-06-21
[#30] Fix tab dot: parked tabs always red due to branch-mismatch check

tejassinghbhati:
parked_tab_status was comparing the parked tab's recorded branch against workspace_branch_name() at render time. When two tabs share the same workspace directory (the common case with Alt+Insert) they read the same .git/HEAD, so any branch change in the active tab immediately turned the parked tab's dot red, even though the parked tab was perfectly healthy. This produced the "alternating by view" pattern Jesper observed: the active tab is always hardcoded Valid (green) while the parked tab ran the branch check and returned BranchGone (red).

The branch a parked tab is on is already shown in the status bar of that tab; the dot does not need to duplicate it. Remove the branch comparison; the dot now only turns red when the workspace directory itself is gone (deleted project, unmounted drive).

### DIFF
--- a/src/bin/orangu/main.rs
+++ b/src/bin/orangu/main.rs
@@ -1497,18 +1497,21 @@ async fn run() -> Result<()> {
 }
 
 /// Determine the feedback-dot status of a parked (non-active) workspace tab.
-/// Checks whether the tab's recorded branch is still the workspace's live
-/// branch; a mismatch signals that the branch was deleted or merged.
+///
+/// The dot is green (Valid) when the workspace directory still exists and no
+/// background LLM task is running, and red (BranchGone) only when the
+/// directory itself has disappeared (deleted project, unmounted drive, …).
+///
+/// A branch-mismatch check was tried here but produced false positives: two
+/// tabs that share the same workspace path read the same `.git/HEAD`, so any
+/// branch change in one tab incorrectly turns the other tab red. Branch state
+/// is already visible in each tab's status bar, so the dot does not need to
+/// duplicate it.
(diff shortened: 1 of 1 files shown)


tejassinghbhati:
@jesperpedersen The alternating red/green bug is fixed in this PR root cause explained above. One thing worth discussing before merge though:

**Dot semantics**

You described the dots as:
- 🟢 = access to llama.cpp
- 🔴 = no access to llama.cpp
- Blinking ⚪ = working

The server is shared across all tabs, so per-tab server connectivity isn't something we can express — if the server goes down, every tab is affected equally. The current implementation uses the dot to mean:

- 🟢 = workspace directory exists
- 🔴 = workspace directory is gone (deleted project, unmounted drive)
- Blinking ⚪ = background LLM response running in this tab

Happy to adjust if you had something different in mind. One thing that *would* be genuinely useful per-tab is a **"has unread output"** state a distinct dot when a parked tab's background response has finished but the user hasn't switched back to read it yet. That's currently a gap: there's no signal that a parked tab is done streaming. Should I add that
as a follow-up to this PR or open a separate issue?

jesperpedersen:
Rebased and merged.

jesperpedersen:
@tejassinghbhati Each workspace can use a different server/model/workspace. Think about WS1 doing coding, while WS2 is doing auto review -- each WS use a different server on the network

jesperpedersen:
@tejassinghbhati Well, we should watch the output for each workspace and update its output window / auto review window. So, its blinking :white_circle: to :green_circle: "automatically" even though you are work in another workspace

jesperpedersen:
@tejassinghbhati Do that as 2 separate PRs

PR #126, opened 2026-06-21
[#30] Real-time parked-tab dot: blinking white -> green on stream finish

tejassinghbhati:
Parked tabs now transition from blinking white to green automatically when their background LLM task finishes, without requiring a tab switch.

Previously the dot color was computed once per main-loop iteration and frozen for the entire wait. Now each render tick inside drive_handle, wait_for_local_command and wait_for_streaming_command calls dot_status() on each parked tab's live JoinHandle via is_finished(), so the dot updates in real time as soon as the task completes.

- WorkspaceTab::dot_status(): canonical per-tab dot logic, checks handle.is_finished() to distinguish Working from Valid
- WaitContext::parked_tabs: slice passed through all three wait fns
- live_tab_statuses(): recomputes fresh Vec<TabStatus> from parked tabs on every render tick inside any wait
- parked_tab_status() removed; callers use dot_status() directly

### DIFF
--- a/src/bin/orangu/input.rs
+++ b/src/bin/orangu/input.rs
@@ -268,6 +268,7 @@ pub struct WaitContext<'a> {
     /// wait (streaming, blocking command), the action is stored here instead of
     /// dropped, so `main` can apply it as soon as the wait returns.
     pub deferred_tab: &'a mut Option<crate::workspace_tab::TabAction>,
+    pub parked_tabs: &'a [crate::workspace_tab::WorkspaceTab],
 }
 
 #[derive(Default)]
--- a/src/bin/orangu/main.rs
+++ b/src/bin/orangu/main.rs
@@ -473,8 +473,8 @@ async fn run() -> Result<()> {
             placement: config.workspaces,
         });
         // Per-tab colored dots: only computed when feedback is on and multiple
-        // tabs are open. Indexed left-to-right; parked tabs check whether their
-        // recorded branch is still the workspace's live branch.
+        // tabs are open. The wait loop recomputes these from live handle state
+        // on every render tick so parked-tab dots update without a tab switch.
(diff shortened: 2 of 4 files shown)


jesperpedersen:
Merged.

jesperpedersen:
@tejassinghbhati `/workspace <DIR>` creates a new workspace with that `<DIR>`.

Since we have Alt+INSERT now, it  should just switch the active workspace to `<DIR>` instead

jesperpedersen:
@tejassinghbhati We should have `create workspace <DIR>` and `delete workspace` for Alt+INSERT and Alt+DELETE. Latter is the hardest as we have `delete <BRANCH>` already

jesperpedersen:
@tejassinghbhati `~/.orangu/workspaces` doesn't look correct - both using a single workspace, and multiple

jesperpedersen:
@tejassinghbhati So, test `-a` as well

tejassinghbhati:
@jesperpedersen Thanks for the merged PR and the follow-up notes.

On `/workspace <DIR>` — agreed, switching the active tab's workspace makes more sense now that Alt+INSERT handles new tab creation. I'll change the handler to switch in place rather than open a new tab.

On `create workspace <DIR>` / `delete workspace` — I'll add those as explicit command variants. For `delete workspace` I'll make sure the parser checks for the `workspace` keyword before falling through to the existing `delete <BRANCH>` path to avoid ambiguity.

On `~/.orangu/workspaces` — I'll look into that now. If both tabs are showing the same single workspace it sounds like the save path is only writing the active tab's workspace and not the full ring. I'll trace through `session_store` and report back.

jesperpedersen:
@tejassinghbhati We likely need the session identifier in `~/.orangu/workspaces` too

Issue #130, opened 2026-06-23
Compressor layer


jesperpedersen:
Add an optional token compression layer in order to limit interaction with the LLM model.

Look at

- [x] https://github.com/mnemosyne-systems/orangu/issues/39
- [x] https://github.com/mnemosyne-systems/orangu/issues/52
- [x] https://github.com/mnemosyne-systems/orangu/issues/62
- [x] https://github.com/mnemosyne-systems/orangu/issues/115

for ideas

PR #134, opened 2026-06-25
[#130] Track internal context compression savings in /usage stats

sksingh2005:


jesperpedersen:
## Overall

- Mixed readiness: Core logic for compression stats and interactive review is mostly complete but hampered by several rejections.
- High risk: Significant regressions in test coverage, O(N*M) performance bottlenecks, and TOCTOU race conditions in configuration.
- Common theme: Integration of `CompressionStats` across the toolchain, including telemetry, UI, and LLM context preparation.
- Common theme: Inconsistent documentation and missing unit tests for new public functions and refactored logic.

## Code

- **src/bin/orangu/completion/mod.rs:423-430**: O(N*M) complexity: `SlashCommand::iter()` is called inside a loop over `skills.all()`, causing redundant iterations.
- **src/bin/orangu/completion/mod.rs:431-436**: Logic error: The check `!SlashCommand::iter().any(...)` inside the loop repeats the full iteration of commands for every single skill.
- **src/bin/orangu/dispatch.rs:392-394**: None
- **src/bin/orangu/dispatch.rs:418-420**: None
- **src/bin/orangu/dispatch.rs:462-464**: None
- **src/bin/orangu/main.rs:1335**: [90] The `if let Ok` pattern silently ignores failures to acquire the `compression_metrics` lock; consider logging the error if locking is critical for telemetry.
- **src/bin/orangu/review/auto.rs:1632-1634**: The mutex lock is handled with `if let Ok(mut metrics)`, which is safe but could technically fail if the mutex is poisoned; however, for a metrics collection use case, this is an acceptable way to silently skip recording if the system is in an inconsistent state.
- **src/bin/orangu/review/interactive.rs:280**: Function name change and return type update are consistent with the logic added in the body.
- **src/bin/orangu/review/interactive.rs:298-302**: Correctly handles the new tuple return from `prepare_llm_diff_context_with_stats`.
- **src/bin/orangu/review/interactive.rs:308-312**: Correctly constructs the final return tuple including the new `stats` variable.
- **src/bin/orangu/stats.rs:113**: None
- **src/compression.rs:332-338**: The `original_lines` count is set to 0 even if `output` contains text, which may lead to inaccurate telemetry for non-matching grep results.
- **src/config.rs:458**: [90] `home::home_dir()` returns an `Option`, but `unwrap_or_default()` on a `PathBuf` is not standard; if it fails, it may result in an empty path rather than a safe fallback.
- **src/config.rs:464-467**: [85] The logic uses `exists()` followed by `read_to_string()`, which is susceptible to a TOCTOU (Time-of-Check to Time-of-Use) race condition.
- **src/config.rs:481-486**: [70] Hardcoding a specific Markdown structure with `<INSTRUCTIONS>` tags might conflict with other LLM prompt engineering templates or specific parser expectations.
- **src/tools.rs:1007**: None
- **src/tools.rs:1008**: None

## Security

No issues found

## Memory

No issues found

## Performance

- **src/bin/orangu/completion/mod.rs:415-434**: O(N*M) complexity: `SlashCommand::iter()` is called repeatedly inside a loop over `skills.all()`, leading to redundant iterations and inefficient lookups.
- **src/compression.rs:252**: `diff.lines().count()` and `content.lines().count()` are called on the full string before compression; for very large inputs, this is an O(n) operation that could be avoided if only the final compressed count is required.
- **src/config.rs:458-461**: Multiple calls to `std::fs::read_to_string` and `exists()` result in redundant system calls; using `std::fs::read_to_string` directly and handling the `Err` case would be more efficient.

## Test Suite

- **src/bin/orangu/completion/mod.rs:411-434**: The new function `slash_command_dropdown_candidates` is added but the existing test code for this function was deleted at the bottom of the file, resulting in zero test coverage for the new logic.
- **src/bin/orangu/main.rs:1333**: No unit tests exist to verify that `metrics.record(&stats)` correctly updates the shared `compression_metrics` state.
- **src/bin/orangu/main.rs:1442**: The addition of `tools.compression_metrics.clone()` to the `skills` initialization lacks a corresponding test to ensure the clone maintains the correct shared reference or state.
- **src/bin/orangu/review/interactive.rs:280**: Renaming the function signature and changing the return type to a tuple will break all existing calls to `build_review_prompt` in the codebase.
- **src/bin/orangu/review/interactive.rs:295-302**: The new return type `(String, orangu::compression::CompressionStats)` requires updated unit tests to verify the `CompressionStats` values are correctly populated.
- **src/config.rs:459-461**: The `home_dir()` call is wrapped in a `map` but the `unwrap_or_default()` on a `PathBuf` is not standard; if `home_dir()` returns `None`, it may lead to unexpected behavior or logic errors in path joining.
- **src/config.rs:482-492**: No unit tests were added to verify the logic of `load_agents_instructions`, specifically for cases where files are missing, empty, or contain special characters.
- **src/config.rs:741-773**: The removal of the original `load_agents_instructions` function without ensuring the new implementation is fully covered by the existing test suite (or adding new tests) creates a regression in test coverage.
- **src/tools.rs:1007**: None
- **src/tools.rs:1008**: None

## Documentation

- **src/bin/orangu/completion/mod.rs:409-412**: Missing doc comments for the public function `slash_command_dropdown_candidates`.
- **src/bin/orangu/review/interactive.rs:280**: The function name was changed to `build_review_prompt_with_stats`, but the docstring (not shown in diff but implied by the change) was not updated to reflect the new return type or the inclusion of `CompressionStats`.
- **src/compression.rs:251-256**: Missing doc comments for the new `prepare_llm_diff_context_with_stats` function.
- **src/compression.rs:261-266**: Missing doc comments for the new `prepare_llm_file_context_with_stats` function.
- **src/compression.rs:315-320**: Missing doc comments for the new `prepare_llm_grep_context_with_stats` function.
- **src/config.rs:458**: [90] Missing doc comment for public function `load_agents_instructions` explaining its purpose and parameters.
- **src/config.rs:458-497**: [80] Lack of internal comments explaining the priority or logic of merging global and local agent instructions.

## Conclusion

- Rejected: **src/bin/orangu/completion/mod.rs**
- Rejected: **src/bin/orangu/review/interactive.rs**
- Rejected: **src/config.rs**

Generated by: **[orangu](https://mnemosyne-systems.github.io/orangu/) 0.9.0** (bartowski/gemma-4-12B-it-GGUF)

jesperpedersen:
```
Context Cache:
Total reads      : 0
Cache hits       : 0
Cache misses     : 0
Cache rate       : 0.0%
Bytes saved      : 0.0 KB

Context Compression:
Original lines   : 4201
Compressed lines : 4159
Lines saved      : 42 (1.0%)
Patterns applied :
  - git_diff (61x)
```

jesperpedersen:
@sksingh2005 This looks a little weird...

jesperpedersen:
@sksingh2005 1.0% for auto review... This should be the biggest saving we can get

jesperpedersen:
Rebased and merged.

sksingh2005:
Sorry one change was remaining I forgot to push :(

jesperpedersen:
@sksingh2005 Just create a new PR - we havn't released yet

jesperpedersen:
@sksingh2005 PTAL me once you are happy

Issue #140, opened 2026-06-26
Reverse compression


sksingh2005:
Introduce a "reverse compression" mechanism that allows the LLM to retrieve the full, uncompressed text of files, diffs, and grep outputs on demand. Currently, when orangu compressed large text to fit within context limits, the LLM has no way to view the omitted sections if it needed more detail.

PR #141, opened 2026-06-26
[#140] Optimize and secure reverse compression

sksingh2005:


jesperpedersen:
## Overall

- This change set introduces a new compression caching system and updates the CLI tool's dispatch and review logic.
- Significant security risks exist in the cache implementation, specifically regarding path traversal and high hash collision probability.
- Performance and stability concerns are present due to non-atomic file operations and potential OOM issues with large file reads.
- Several components are rejected or require rework due to missing unit tests, inconsistent patterns, and initialization order errors.
- Documentation is incomplete across the new compression modules and interactive review prompts.

## Code

- **src/bin/orangu/main.rs:1336**: Passing a reference to a value inside a `Mutex` or `RwLock` (implied by `tools.compression_store`) might lead to lifetime issues or unnecessary complexity if the receiver expects an owned value.
- **src/bin/orangu/main.rs:1451**: Inconsistent pattern: the first change passes a reference (`Some(&*...)`), while the second change passes a clone (`.clone()`). This may cause type mismatches in the receiving function.
- **src/bin/orangu/workspace_tab.rs:112-118**: `ToolExecutor::with_config` is called before `std::fs::create_dir_all(&session_dir)`, which may cause the tool executor to fail if it attempts to access or initialize the directory before it exists.
- **src/compression_cache.rs:47**: [90] Potential for hash collisions using only the first 10 characters of a SHA-256 hash in a high-volume cache.
- **src/compression_cache.rs:55**: [85] Race condition: `path.exists()` check followed by `std::fs::write` is not atomic; use `std::fs::OpenOptions` with `.create_new(true)`.
- **src/compression_cache.rs:72**: [80] Path traversal check is incomplete; it doesn't prevent `..` or other relative path segments.
- **src/compression_cache.rs:74**: [75] `cache_dir` is accessed via `self.cache_dir` but the code uses `dir` (which is shadowed/local) or incorrect variable naming in logic.
- **src/tools.rs:122-125**: [90] `compression_store` is initialized using `session_dir.clone()`, which is correct, but the `session_dir` field is also assigned the original `Option<PathBuf>`, ensuring consistency between the store's path and the executor's field.

## Security

- **src/compression_cache.rs:47-48**: Potential hash collision: Using only the first 10 characters of a Sha256 hash significantly increases the probability of collisions, which could lead to one user's content overwriting or being retrieved as another's.
- **src/compression_cache.rs:62-63**: Path Traversal Risk: While there is a basic check for `/`, `\`, and `.`, it does not prevent all forms of path traversal or platform-specific reserved names, which could allow access to files outside the `compression_cache` directory.

## Memory

No issues found

## Performance

- **src/bin/orangu/workspace_tab.rs:116-121**: Moving `ToolExecutor::with_config` and `SkillRegistry::discover` after `session_dir_path` and before `create_dir_all` causes redundant initialization or potential failures if the directory doesn't exist yet.
- **src/compression_cache.rs:46-47**: Using a 10-character substring of a SHA-256 hash increases the risk of collisions, which could lead to data corruption if two different contents produce the same short hash.
- **src/compression_cache.rs:74-75**: `std::fs::read_to_string` reads the entire file into memory. For very large cached contents, this could lead to high memory pressure or OOM errors.

## Test Suite

- **src/bin/orangu/review/interactive.rs:289**: Missing unit tests for the new `store` parameter in `build_review_prompt_with_stats` to ensure correct handling of `None` and `Some` cases.
- **src/bin/orangu/review/interactive.rs:304**: Lack of regression tests for the internal `context` construction when `store` is present to verify it doesn't interfere with existing prompt logic.
- **src/bin/orangu/workspace_tab.rs:112-118**: Moving `ToolExecutor` and `SkillRegistry` initialization after `session_dir` path resolution ensures that the executor has access to the specific session directory, but this change requires ensuring that existing unit tests for `WorkspaceTab` initialization are updated to account for the new `session_dir` dependency.
- **src/compression_cache.rs:1-82**: No unit tests are provided for the new `CompressionStore` struct to verify hashing, file I/O, and error handling.
- **src/compression_cache.rs:41-43**: Potential hash collision risk: using only the first 10 characters of a SHA256 hash significantly increases the probability of collisions compared to using the full hash.

## Documentation

- **src/bin/orangu/review/interactive.rs:286**: The new parameter `store` lacks a doc comment explaining its purpose or usage within the prompt construction.
- **src/bin/orangu/review/interactive.rs:300**: The call site for `build_review_prompt_with_stats` does not include a comment clarifying how the `store` option influences the output.
- **src/compression_cache.rs:19-21**: Missing doc comments for the `CompressionStore` struct explaining its purpose as a persistent or session-based cache.
- **src/compression_cache.rs:31-32**: The doc comment for `store` mentions "full content," but doesn't specify that it returns a truncated 10-character hash for filename construction.
- **src/compression_cache.rs:54-55**: The `retrieve` method lacks a doc comment explaining its parameters and expected return values.

## Conclusion

- Rejected: **src/bin/orangu/review/interactive.rs**
- Rejected: **src/bin/orangu/workspace_tab.rs**
- Rejected: **src/compression_cache.rs**

Generated by: **[orangu](https://mnemosyne-systems.github.io/orangu/) 0.9.0** (bartowski/gemma-4-12B-it-GGUF)

jesperpedersen:
Do we need to tie this into the compression statistics ?

jesperpedersen:
@sksingh2005 Please, rebase

sksingh2005:
> Do we need to tie this into the compression statistics ?

 It is actually already fully tied in

jesperpedersen:
Rebased and merged.

Issue #145, opened 2026-06-26
Per-workspace LLM configuration


tejassinghbhati:
### Per-Workspace `/server` and `/model`
`/server` and `/model` are currently global across all open workspace tabs. Changing the model in one tab immediately affects every other tab, making it difficult to use different models for different projects.
For example, a developer may want to use a small, fast model for day-to-day editing in a large monorepo, while reserving a larger reasoning model for code review or debugging tasks in another tab.

The groundwork for this already exists. `workspace_tab.rs` currently contains the following note:

> The model and server are shared across tabs (per-workspace `/server`/`/model` come later).

This issue tracks the implementation of that "come later."
### Problem

Consider the following multi-tab workflow:

| Tab | Workspace | Ideal Model |
|-----|------------|-------------|
| 1 | `~/projects/backend` | `gemma-4-12B` — fast, good for file edits |
| 2 | `~/projects/frontend` | `gemma-4-27B` — better at TypeScript |
| 3 | `~/projects/backend` (review branch) | `qwen-72B` — slower but more thorough for `/auto_review` |

Currently, switching to the 72B model in tab 3 silently changes the model used by tabs 1 and 2 as well. There is no way to pin a model or server to a specific workspace.

### Proposed Solution

The fix has two parts. First, `current_server` and `current_model` should move from global run-loop locals into `WorkspaceTab` so that `/server` and `/model` changes in one tab are isolated to that tab. The tab switch macros `load_tab!` and `save_all_tabs!` would restore and persist these alongside the existing session ID. Second, orangu should look for an optional `.orangu/orangu.conf` inside the workspace directory and merge it on top of the global `~/.orangu/orangu.conf` when the tab is opened, so a repo can pin its own server and model without touching the global config. The layering order would be: `global config` → `workspace config` → `runtime /server//model` commands (all tab-local).

The change touches `workspace_tab.rs` to add the new fields and wire them through the tab switch `macros`, `main.rs` to thread the per-tab `server/model` into `ToolExecutor` on each load, config.rs to add a `load_workspace_configuration` function that merges on top of the global config, and optionally `session_store.rs` to persist the chosen server/model in `~/.orangu/workspaces`. The manual pages for configuration and workspaces would be updated to document the new layering and the fact that `/server` and `/model` are now tab-local.

[...]

PR #146, opened 2026-06-26
[#145] Per-workspace LLM configuration: isolate /server and /model per tab

tejassinghbhati:
Isolates `/server` and `/model` per workspace tab so changing them in one tab no longer affects others. Adds optional `.orangu/orangu.conf` support inside any workspace directory to pin a server and model for that project. Tab switches now restore the correct server and model for the incoming tab.


doc/manual/en/20-configuration.md
```diff
@@ -62,6 +62,38 @@ available after `--init`:
...
+```
+<workspace>/.orangu/orangu.conf
+```
+
+Only the `[orangu]` section's `server` and `model` keys are read from this file;
+all other settings (token limits, compression, platform, etc.) remain global.
+The layering order is:
+
+1. `~/.orangu/orangu.conf` — global defaults
+2. `<workspace>/.orangu/orangu.conf` — workspace override
```
jesperpedersen: We should stick this in the session directory - maybe settings, so

```
~/.orangu/sessions/<UUID>/settings
```

tejassinghbhati:
storing it in `~/.orangu/sessions/<UUID>/settings` ties the server/model choice to the session rather than the workspace directory, which means it persists across restarts automatically and doesn't require the user to create any files manually. 
-- This would be a cleaner approach. Thankyou so much for the review
Sure I'll make the changes.

src/config.rs
```diff
@@ -452,6 +452,8 @@ fn normalize_llm_configuration(llm: &mut LlmConfiguration) -> Result<()> {
     }
 }
 
+/// Read `<workspace>/.orangu/orangu.conf` and extract the `[orangu]` section's
```
jesperpedersen: This is wrong

jesperpedersen:
Merged.

PR #148, opened 2026-06-30
Fix docs publish pipeline: PAT auth, link references, and image assets

tejassinghbhati:
Repairs the `publish-docs` workflow so the website actually updates on tag push, and so the generated pages are correct.

## Why

The job failed on tag push with:

```
fatal: unable to access 'https://!fdRE$#hgYT56@github.com/.../mnemosyne-systems.github.io.git/': URL rejected: Bad hostname
```

The `WEBSITE_PAT` was inlined into the clone URL, so URL-significant characters in the token (`#`, `@`, `/`) corrupted it. Beyond that, the generated pages had two long-standing issues: reference-style links rendered as raw `[text][ref]`, and the introduction screenshot overflowed the column / was not copied to the site.

## Changes

- **Token via auth header.** Switch the website push to `actions/checkout` with `token:`, which passes the PAT as a base64 `Authorization` header instead of embedding it in the URL — immune to special characters.
- **Resolve link references.** Pass `doc/manual/en/99-references.md` as a second input to every `pandoc` conversion so reference-style links resolve.
- **Image sizing.** Restore the image-sizing `<style>` block in `web-template.html` (caps the terminal screenshot at 640px).
- **Copy image assets.** Copy `doc/images/*.png` into the site's `doc/images/`.

> Note: this still requires `WEBSITE_PAT` to be set to a valid GitHub token with write access to the website repo — the current secret value is not a real token.

Companion website PR with the regenerated 0.10.0 docs: mnemosyne-systems/mnemosyne-systems.github.io#9

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -23,12 +23,18 @@ jobs:
           mkdir -p target/web-doc
           MANUAL=doc/manual/en
           TMPL=doc/manual/web-template.html
+          # The link reference definitions ([orangu], [llama], community links,
+          # etc.) live in 99-references.md. Pandoc resolves reference-style links
+          # only when that file is supplied alongside each page, so it is passed
+          # as a second input to every conversion; otherwise the links render as
+          # raw "[text][ref]" text.
+          REFS="$MANUAL/99-references.md"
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

Issue #133, opened 2026-06-24
Knowledge graph


sksingh2005:
https://github.com/safishamsi/graphify

PR #149, opened 2026-06-30
[#133] Knowledge graph

sksingh2005:


jesperpedersen:
@sksingh2005 Remember license header in all new files

jesperpedersen:
@sksingh2005 See how many languages you can support through `tree-sitter`

sksingh2005:
@jesperpedersen for /export graph is it needed because exporting it into pdf will loose interactivity as it will just turn it into a frozen snapshot. Html file is easily shareable and while opening html it can be exported to pdf. 
Let me know your thoughts!

jesperpedersen:
@sksingh2005 Yeah, we can start with HTML. But, the code duplication report has internal and external links in the PDF so it can be done

jesperpedersen:
@sksingh2005 The branch is called `knowledge-graph`, so it should be

```
orangu-knowledge-graph-graph.html
```

And, we should make the report name a link in the output window so you can click on it. That way we can remove "Open that file in your browser to explore the graph."

jesperpedersen:
@sksingh2005 Maybe we should find the main function and make that file the default view

jesperpedersen:
@sksingh2005 Lets skip PDF for this

jesperpedersen:
@sksingh2005 clippy

doc/manual/en/41-core_tools.md
```diff
@@ -868,3 +868,36 @@ export review
...
 export duplicates
 ```
+
+\newpage
+
+## /graph
+
+Generates an interactive, standalone HTML visualization of the workspace's Knowledge Graph.
+
+Orangu incrementally extracts a semantic graph of the codebase in the background using Tree-sitter. This command takes that internal graph and writes it out to `orangu-graph-graph.html` using the `vis-network` JavaScript library.
```
jesperpedersen: We have to explicit about that it is

```
<REPOSITORY_NAME>-<BRANCH_NAME>-graph.html
```

src/bin/dump_graph.rs
```diff
@@ -0,0 +1,11 @@
+use orangu::agents::hooks::run_session_start_hook;
```
jesperpedersen: License header

src/bin/dump_graph.rs
```diff
@@ -0,0 +1,11 @@
...
+use std::path::Path;
```
jesperpedersen: Do we need this file ?

src/bin/dump_graph.rs
```diff
@@ -0,0 +1,11 @@
...
+
+fn main() {
+    println!("Scanning Orangu workspace...");
+    let result = run_session_start_hook(Path::new("."));
+    println!("{}", result.summary());
+
+    // The cache is also automatically saved to .orangu/kg_cache.json by the hook.
+    println!("Done!");
```
jesperpedersen: We don't need this

src/bin/dump_graph.rs
sksingh2005: No no not needed. Sorry it got accidentally pushed :)

src/bin/dump_graph.rs
sksingh2005: This file got accidentally pushed

doc/manual/en/41-core_tools.md
```diff
@@ -868,3 +868,36 @@ export review
...
 export duplicates
 ```
+
+\newpage
+
+## /graph
+
+Generates an interactive, standalone HTML visualization of the workspace's Knowledge Graph.
+
+Orangu incrementally extracts a semantic graph of the codebase in the background using Tree-sitter. This command takes that internal graph and writes it out to `orangu-<branch>-graph.html` (for example, `orangu-knowledge-graph-graph.html` when on the `knowledge-graph` branch) using the `vis-network` JavaScript library.
```
jesperpedersen:
```
<repository>-<branch>-graph.html
```

jesperpedersen:
@sksingh2005 Test the patch on different repositories and branches

sksingh2005:
@jesperpedersen yes I have tested in other projects too. Would also need your feedback :)

jesperpedersen:
@sksingh2005 If I call `/graph` inside pgmoneta the report is still called `orangu-main-graph.html` instead of `pgmoneta-main-graph.html`

jesperpedersen:
@sksingh2005 Maybe we can have a "select all" and "clear all" buttons to make it easier to deal with the file selection

jesperpedersen:
@sksingh2005 Looks like it is using the directory name instead of the repository name. `/graph` and `/exporter` should be able to share the same function for the prefix of the name.

Also, the output from `/graph` isn't rendered as Markdown in the window

sksingh2005:
> @sksingh2005 Looks like it is using the directory name instead of the repository name. `/graph` and `/exporter` should be able to share the same function for the prefix of the name.
> 

@jesperpedersen  Thanks I will be updating like that only

> Also, the output from `/graph` isn't rendered as Markdown in the window

Thanks for pointing that out I will update.

jesperpedersen:
Merged.

PR #153, opened 2026-07-03
[#152] Local semantic code search with hybrid retrieval

### LINKED ISSUE #152: Local semantic code search with hybrid retrieval
### Summary

orangu can retrieve code lexically (`/grep`), by symbol name and structure (the #133 knowledge graph), and by AST shape (`/duplicates`), but it cannot retrieve code by *meaning*. A query like "where is rate-limiting handled?" won't find `throttle_requests()` — the words don't match and graph centrality won't surface it. This adds a semantic retrieval layer, computed fully offline, and fuses it with the existing knowledge graph.

### Proposed solution

Build a local embedding index over the codebase. Reuse the Tree-sitter chunking that already backs `/graph` and `/duplicates` to split the workspace into function- and symbol-level chunks, embed each one through the configured server's `/v1/embeddings` endpoint, and persist the vectors on disk with the same incremental invalidate-on-change approach used by the knowledge graph cache, so the index stays cheap to keep current and never leaves the machine.

Retrieval is hybrid: embed the query, rank chunks by cosine similarity, then expand the top matches along the knowledge graph's call edges — a semantic seed followed by structural expansion — and blend the two signals into a single ranked result. Expose it as a `/search <query>` command, and inject the top-ranked chunks into chat and `/review` context automatically, bounded by the existing compression budget so the context window is never flooded.

Because this depends on an embeddings-capable server, it should degrade gracefully: when no `/v1/embeddings` endpoint is configured, the feature stays dormant and orangu falls back to today's graph-and-grep retrieval. A new `[orangu]` setting names the embeddings server/model, and the manual gains a short chapter describing the index and the `/search` command.

### Acceptance criteria

- [ ] The workspace is embedded offline via the server's `/v1/embeddings` endpoint and the index persists across runs
- [ ] The index updates incrementally as files change, like the knowledge graph cache
- [ ] `/search <query>` returns semantically relevant code, ranked by a hybrid of embedding similarity and graph structure
- [ ] Top results are injected into chat/review context within the compression budget
- [ ] With no embeddings endpoint configured, the feature is dormant and existing retrieval is unaffected

tejassinghbhati:
Closes #152.

Adds `/search` — semantic code search that retrieves by *meaning*, complementing the lexical `/grep` and the structural knowledge graph. A query like `where is rate-limiting handled?` can surface a `throttle_requests` function whose name shares no words with the query.

### Design

A new `src/embeddings/` module:

- **`client.rs`** — a minimal OpenAI-compatible `/v1/embeddings` client, built from the server profile named by the new `[orangu].embeddings` setting (llama.cpp serves this endpoint, so the existing `LlmConfiguration` works unchanged).
- **`chunk.rs`** — reuses the Tree-sitter symbol extraction behind `/graph` and `/duplicates` to split files into per-symbol chunks, sliced by the `L<start>-L<end>` line range the extractor already records. Chunk ids match graph node ids, so hits can be expanded along graph edges.
- **`index.rs`** — a persistent, incremental vector index at `.orangu/embeddings_cache.json`, mirroring the knowledge graph cache: a per-file sha256 map re-embeds only changed files. Hybrid search ranks by cosine similarity, then expands the strongest hits along the knowledge graph's call edges — **semantic seed, structural expansion**.

### Behaviour

- `/search <query>` builds/updates the index and embeds the query **off the UI thread**, holding the graph lock only for the brief synchronous ranking step (never across the embedding build).
- Registered as a slash command with completion and natural-language aliases (`search for ...`, `semantic search ...`).
- **Graceful degradation:** when `[orangu].embeddings` is unset or names a nonexistent server, semantic search stays dormant, `/search` explains how to enable it, and `/grep` + the knowledge graph are unaffected.

### Acceptance criteria

- [x] Workspace embedded offline via `/v1/embeddings`; index persists across runs
- [x] Index updates incrementally as files change, like the knowledge graph cache
- [x] `/search` returns results ranked by a hybrid of embedding similarity and graph structure
- [x] Dormant and non-disruptive when no embeddings endpoint is configured
- [ ] Automatic injection of top results into chat/review context — deferred to a follow-up; this PR lands the index + explicit `/search`, which is the foundation that injection builds on

### Notes

- Unit tests cover cosine similarity, chunk slicing, location parsing, and hybrid ranking. `cargo fmt --check` and `cargo clippy --all-targets -D warnings` are clean.
- ~900 lines; the flagship retrieval feature described in #152.

Docs: a `/search` chapter in the Core Tools manual and an `embeddings` row in the configuration table.


doc/etc/orangu.conf
```diff
@@ -6,6 +6,7 @@ max_tool_rounds = 10
 # review_max_tokens = 512
 # code_max_tokens = 0
 # compile_workers = 0    # parallel /build jobs, e.g. `make -j <n>`; 0 = unused
+# embeddings = main-server    # server section for semantic /search; unset = dormant
```
jesperpedersen: This sounds like a server `role` instead - included in `all`

doc/manual/en/20-configuration.md
```diff
@@ -105,6 +105,7 @@ code_max_tokens = 0
 | `review_max_tokens` | No | Response-token cap for each `/auto_review` request. Defaults to `512`; `0` disables the cap. Raise it (e.g. `2048`) when the review model thinks before answering |
 | `code_max_tokens` | No | Response-token cap for normal chat and tool responses. Defaults to `0` (no cap) |
 | `compile_workers` | No | Parallel job count `/build` passes to toolchains that support one (e.g. `make -j`, `meson compile -j`, `cargo --jobs`). Defaults to `0`, meaning unused: no job flag is passed and each toolchain falls back to its own default |
+| `embeddings` | No | Name of the server section used to embed code for semantic `/search`. When unset (the default), semantic search is dormant and retrieval falls back to the knowledge graph and `/grep`. The named server must expose an OpenAI-compatible `/v1/embeddings` endpoint with an embedding model loaded |
```
jesperpedersen: We should query if the endpoint is available upon startup and enable it if it is

doc/manual/en/41-core_tools.md
```diff
@@ -937,3 +937,51 @@ show graph
...
+`where is rate-limiting handled?` can surface a `throttle_requests` function whose
+name shares no words with the query.
+
+```text
+/search <query>
+```
+
+The first time it runs, orangu embeds every extractable symbol in the workspace
+through the server named by the `[orangu].embeddings` setting and stores the
+vectors in `.orangu/embeddings_cache.json`. Subsequent runs re-embed only the
```
jesperpedersen: We should put this in the session directory

jesperpedersen:
@tejassinghbhati Lets start with these minor changes

doc/manual/en/20-configuration.md
```diff
@@ -180,6 +180,7 @@ model = ggml-org/gemma-4-E4B-it-GGUF
 | `endpoint` | Yes | OpenAI-compatible API URL |
 | `model` | No | Model identifier used in chat completion requests. Overrides the general `[orangu].model` when set |
 | `api_key` | No | API key sent as `Authorization: Bearer <key>` on every request to the server. Required when a llama.cpp server runs with `--api-key`, or for any authenticated OpenAI-compatible endpoint |
+| `role` | No | What the server is used for. Defaults to `all` (serves every role). Set `role = embeddings` to designate the server that embeds code for semantic `/search`; an `all` server also serves embeddings. Semantic search is enabled automatically when that server's `/v1/embeddings` endpoint responds at startup |
```
jesperpedersen: This need to be sync'ed with CONFIGURATION.md

tejassinghbhati:
Implemented all three

1. **Server role instead of a key** : dropped `[orangu].embeddings`; embeddings is now resolved by role (`role = embeddings`, or the default `all` server). Documented the role in the server-section tables and the example config.
2. **Auto-detect at startup** : orangu now probes the candidate server's `/v1/embeddings` on startup and enables `/search` only when it responds; a chat-only server fails the probe and search stays dormant.
3. **Session directory** :  the index now lives at `~/.orangu/sessions/<UUID>/embeddings_cache.json`.

One thing worth deciding on #3: storing the index per-session means each new session on the same workspace re-embeds the codebase from scratch, since the cache isn't shared across sessions (the knowledge-graph cache, by contrast, lives in the workspace `.orangu/` and is reused). For an expensive embedding pass that's a fair bit of recompute. If cross-session reuse matters, a workspace-keyed location under `~/.orangu/` (e.g. `~/.orangu/embeddings/<workspace-hash>/`) would keep it out of the workspace tree while still reusing across sessions.
 
Happy to go either way, please let me know which you prefer.

src/bin/orangu/commands/mod.rs
```diff
@@ -353,6 +353,9 @@ pub enum LocalCommand<'a> {
     SetVerbosity(&'a str),
     Diff(Option<Cow<'a, str>>),
     Grep(Option<Cow<'a, str>>),
+    /// `/search <query>`: semantic code search over the embedding index, fused
```
jesperpedersen: Remove the comment

src/bin/orangu/commands/mod.rs
```diff
@@ -429,6 +432,10 @@ pub struct CommandContext<'a> {
     pub tools: &'a ToolExecutor,
     pub workspace: &'a Path,
     pub session_dir: &'a std::path::Path,
+    /// The server section serving embeddings for semantic `/search`, resolved by
```
jesperpedersen: Remove the comment

jesperpedersen:
@tejassinghbhati We can do the workspace cache if we can make sure that it is up-to-date

jesperpedersen:
@tejassinghbhati Let me know when you are happy with it

tejassinghbhati:
Moved it to the workspace cache (`.orangu/embeddings_cache.json`, alongside `kg_cache.json`). It's kept up-to-date automatically. Pushed.

jesperpedersen:
@tejassinghbhati I meant `~/.orangu/embeddings/workspace/<HASH>/cache.json`

tejassinghbhati:
Apologies for that. the index now lives at `~/.orangu/embeddings/workspace/<sha256(workspace-path)>/cache.json`.

jesperpedersen:
@tejassinghbhati We can delete this "Semantic /search enabled via main-server"

jesperpedersen:
@tejassinghbhati We should document an embedding server so with

```
 --embedding \
 --pooling last
```

jesperpedersen:
@tejassinghbhati I did `/search security` after startup - killed after 2h. `<ESC> <ESC>` didn't work

jesperpedersen:
@tejassinghbhati Did you track it down ?

tejassinghbhati:
yes, I apologize, it is fixed now. 
`/search` now streams progress and is cancellable with double-Esc. 

Root cause: it ran as an un-abortable `spawn_blocking` task and the wait loop never checked for Esc. It's now a streaming command with cooperative cancellation (a flag the build checks between files), and the indexing pass is resumable, a cancelled build saves what it embedded and the next `/search` continues from there rather than restarting. Pushed.

jesperpedersen:
@tejassinghbhati How long does it take on your machine ?

jesperpedersen:
@tejassinghbhati Rebase

jesperpedersen:
@tejassinghbhati instead of

```
Embedded X/Y files...
```

we can do

```
Working (3%) (4m52s)
```

jesperpedersen:
@tejassinghbhati Can we calculate a remaining time ?

jesperpedersen:
@tejassinghbhati Its been working for 10m and its at 7/82... so having an estimate would be good... maybe calculate based on the size of the files

jesperpedersen:
@tejassinghbhati I wonder if we can do all the calculations first for all files, and then send it - so 2 x 82 steps. But, maybe the first phase can use `compile_workers` to do the work in parallel. 2nd phase will be a sequence

jesperpedersen:
@tejassinghbhati It is still working, but there is still no data in `~/.orangu/` of the files that has been processed

jesperpedersen:
@tejassinghbhati We need to create the cache file first, and then append to it when processing the next file

jesperpedersen:
@tejassinghbhati We can stick another file in the directory that has the path to the file, and its timestamp when it was processed

jesperpedersen:
@tejassinghbhati If `compile_workers` will work then we should take a value of `0` as use all CPU threads, so we will need to get that information

tejassinghbhati:
Pushed the latest round: two-phase build (parallel parse via compile_workers, 0 = all CPU threads; sequential embed), so the ETA is accurate from the start and shown as Working (7%) (12s, ~9m30s left). The cache is now written incrementally with a processed.log manifest, so progress persists on disk even through a hard kill. Rebased on main, CI green.

jesperpedersen:
@tejassinghbhati How do we know when the cache has to be recalculated, and which part of the cache ? Since if a file hasn't changed then we can keep its current entries

jesperpedersen:
@tejassinghbhati We need to look at remaining time. It keeps counting up - it only counts down when a file has been processed

jesperpedersen:
@tejassinghbhati We have to look at remaining time, as it only count up. It count down when a file has been processed - however, I see interaction with the LLM...

jesperpedersen:
@tejassinghbhati The 1st phase should be 50%, and the 2nd phase the other 50%

jesperpedersen:
@tejassinghbhati It still crashes my computer

jesperpedersen:
@tejassinghbhati Processing the files take no time on my machine - so it all the upload to the LLM. We need to see we can optimize that part

jesperpedersen:
@tejassinghbhati Remaining time is still buggy `Working (50%) (6m3s, ~0s left)`

jesperpedersen:
@tejassinghbhati `Working (51%) (6m17, ~4m37s left)`

jesperpedersen:
@tejassinghbhati I think we need to expand the `processed.log` because it says `50%` but the log says

```
1783159422	src/agents/mod.rs
1783159452	src/context/fragments.rs
1783159524	src/llm/mod.rs
```

jesperpedersen:
@tejassinghbhati And, remaining still make massive jumps

jesperpedersen:
@tejassinghbhati Why `chunks.jsonl` ? Shouldn't it be `chunks.json` ?

jesperpedersen:
@tejassinghbhati `meta.json` only contains a few of the files... We should do all the local files first, and then upload after

jesperpedersen:
@tejassinghbhati `<ESC><ESC>` just says "Cancelling...` - it doesn't stop

jesperpedersen:
@tejassinghbhati Then

```
Search cancelled after 16m40s.
:green_circle: 
```

Remove "Cancelling..." and tt should be 

```
Request cancelled.
:red_circle: 
```

tejassinghbhati:
Local-first, then upload. Restructured into two clean phases, the first half (0–50%) does all the local parsing up front and writes the full meta.json (every file) and processed.log, so the on-disk state lists every file before anything is uploaded. The second half (50–100%) is purely the upload.

Faster upload. Embedding requests are now pipelined, up to 8 files in flight at once, so the server (the real bottleneck) stays busy instead of idling between round-trips. With -np N on llama-server they run in true parallel.

Cancellation actually stops now. <Esc><Esc> aborts the in-flight requests immediately (was waiting for them to drain before). Removed the "Cancelling…" text; the result is now Request cancelled. with a red dot.

Smaller fixes: chunks.jsonl → chunks.json; the remaining-time estimate is steadier (overall rate, and it counts down between updates). Though with concurrent bursts it can still step a little when several files land at once.

One heads-up: this changes the on-disk cache format, so the first /search after updating does a fresh index build.

jesperpedersen:
@tejassinghbhati I have

```
1783163934	src/graph/store.rs
1783163934	src/bin/orangu/models.rs
17831639341783163934		src/graph/extract.rssrc/bin/orangu/completion/git_refs.rs

1783163934	src/compression_cache.rs
1783163934	src/bin/orangu/commands/natural.rs
```

so there must be a race somewhere

doc/manual/en/73-openai.md
```diff
@@ -128,3 +128,21 @@ llama-server -hf bartowski/Qwen_Qwen3.6-35B-A3B-GGUF \
...
+`--pooling last` (the pooling strategy the model expects):
+
+```sh
+llama-server -hf ggml-org/embeddinggemma-300M-GGUF \
+             --port 8300 \
+             --embedding \
+             --pooling last \
+             --ctx-size 8192 \
+             --fit on
+```
```
jesperpedersen, line 150: We need to explain `-np N` and that if the LLM is restarted then we need to upload again

doc/manual/en/73-openai.md
```diff
@@ -128,3 +128,21 @@ llama-server -hf bartowski/Qwen_Qwen3.6-35B-A3B-GGUF \
...
+
```
jesperpedersen, line 151: Start this, and then start orangu - nothing is detected. And, `/search` doesn't work

doc/manual/en/73-openai.md
```diff
@@ -128,3 +128,21 @@ llama-server -hf bartowski/Qwen_Qwen3.6-35B-A3B-GGUF \
...
+Embedding model
+
+Semantic `/search` needs a server serving an embedding model. Start one with
+`--embedding` (switches the server to the embeddings endpoint) and
+`--pooling last` (the pooling strategy the model expects):
+
+```sh
+llama-server -hf ggml-org/embeddinggemma-300M-GGUF \
+             --port 8300 \
```
jesperpedersen: Ah, I used the wrong port number due to this - we should use 8100

doc/manual/en/73-openai.md
```diff
@@ -128,3 +128,21 @@ llama-server -hf bartowski/Qwen_Qwen3.6-35B-A3B-GGUF \
...
+             --embedding \
+             --pooling last \
+             --ctx-size 8192 \
```
jesperpedersen, line 145:
```
> /search security                                                                                                                                                                                                                                                                                    
Error: embeddings request failed (status 500 Internal Server Error): {"error":{"code":500,"message":"input (762 tokens) is too large to process. increase the physical batch size (current batch size: 512)","type":"server_error"}}
```

so we need to provide better defaults

tejassinghbhati:
Screenshot 2026-07-04 194447

Confirms the ranking is working correctly.

doc/manual/en/73-openai.md
```diff
@@ -128,3 +128,47 @@ llama-server -hf bartowski/Qwen_Qwen3.6-35B-A3B-GGUF \
...
+(the pooling strategy the model expects), and `-np N` (the number of requests the
+server processes in parallel):
+
+```sh
+llama-server -hf ggml-org/embeddinggemma-300M-GGUF \
+             --port 8100 \
+             --embedding \
+             --pooling last \
+             --ctx-size 8192 \
+             -np 8 \
```
jesperpedersen, line 146: I need to add

```
 -b 2048 \
 -ub 2048 \
```

jesperpedersen:
Merged.

PR #156, opened 2026-07-06
Move knowledge graph cache out of the workspace tree

tejassinghbhati:
Fixes the bug @jesperpedersen reported: `.orangu/kg_cache.json` was being written inside every workspace directory, cluttering the project tree with a file that could accidentally get committed if `.orangu/` isn't gitignored.

## Fix

Both the knowledge graph cache and the embeddings index are **workspace-scoped, not session-scoped** — `run_session_start_hook` and `EmbeddingIndex::build_or_update` both take a workspace path and are shared across every session/tab open on it. So they now live together under one global, per-workspace root, per Jesper's suggested layout:

```
~/.orangu/workspace/<sha256(canonical path)>/graph/
~/.orangu/workspace/<sha256(canonical path)>/embeddings/
```

- New `src/workspace_cache.rs` — the shared hashing/directory helper both subsystems call.
- `GraphCache::cache_path(workspace)` replaces the hardcoded `.orangu/kg_cache.json` path in `run_session_start_hook`.
- The embeddings cache (previously `~/.orangu/embeddings/workspace/<hash>/`, from #152) moves to the new shared layout alongside it, since it was moving anyway.

## Notes

- This changes the on-disk location for both caches, so existing users get a one-time rebuild (the graph re-scans on next session start; the embeddings index re-embeds on next `/search`). No stale `.orangu/kg_cache.json` files are left behind automatically — worth calling out in the changelog if that matters.
- Added unit tests for `workspace_cache_dir`: distinct subdirectories per subsystem, same parent, deterministic, and distinct across workspaces.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.


jesperpedersen:
Merged.

Issue #163, opened 2026-07-08
Coordinator


jesperpedersen:
Create a coordinator that will control which model that is active based on the requests it get.

We need to create a `/v1/coordinator` endpoint for capabilities, and proxy all the llama.cpp endpoint.

The `orangu-coordinator.conf` defines all the models, so orangu just need to connect to the coordinator and it will do the rest

PR #164, opened 2026-07-08
[#163] Add idle timeout, config hot-reloading, fallback routing, and graceful shutdown

sksingh2005:
## Changes
- **Auto-Unload:** Added an optional `idle_timeout` config to automatically unload inactive models and free VRAM.
- **Config Hot-Reloading:** The proxy now periodically polls `orangu-coordinator.conf` and applies routing changes instantly without a restart.
- **Fallback Routing:** Automatically routes to the default `all` profile if a requested model fails to load (e.g. out of memory).
- **Graceful Shutdown API:** Added `GET /v1/coordinator/shutdown` (secured via loopback check and optional `shutdown_token`) to allow clean session termination.
- **Reliability Fixes:** 
  - Orphaned processes are now cleaned up via `.kill_on_drop(true)`.
  - Replaced `SIGKILL` with `SIGINT` to allow `llama-server` to save its KV cache.
  - Fixed multiple edge cases around stale PID kills, fallback credentials, and hot-reload process reconciliation.


jesperpedersen:
@sksingh2005 You are missing the `orangu-coordinator -i` part

jesperpedersen:
Merged.

PR #175, opened 2026-07-09
[#166] New tool: /schedule — cron-style scheduled commands

### LINKED ISSUE #166: New tool: /schedule
Create a new tool called `/schedule` that will execute a job in a cron style manner.

The configuration should be in a crontab like way so you can write

```
0 * * * * /export pr
```

for example

tejassinghbhati:
Closes #166

Adds a crontab-style scheduler. Jobs live in `~/.orangu/schedule`, one per line as five time fields plus the command to run:

```
# hourly pull request report
0 * * * * /export pr
30 6 * * 1-5 /statistics
```

How it works:

- The run loop checks the schedule as it goes around and pushes due commands onto the same pending queue as commands typed mid-request, so a job runs exactly like one the user entered, in the active workspace tab.
- Every minute boundary since the last check is considered, so jobs aren't skipped while the loop is busy; minutes before startup never fire. The file is re-read each minute, so edits apply without a restart.
- Hand-rolled five-field parser, no new dependency: `*`, lists (`1,15`), ranges (`1-5`), steps (`*/10`, `8-18/2`), 0 or 7 for Sunday, and classic cron's either-matches rule when both day fields are restricted. Times are UTC, since orangu carries no timezone database.
- Bare `/schedule` lists the jobs with their next run times, and points out lines that didn't parse instead of silently ignoring them.

Also fixes a small pre-existing parsing bug this feature would have inherited: `show usage` and `show statistics` were being claimed by the `show <file>` natural form as file names. They now resolve to their commands, like `show manual` and `show pending` already did, and `show schedule` gets the same treatment.

Manual chapter, /help text, and unit tests (parser, matching semantics, next-run, crontab splitting) included. Full `cargo test --bin orangu`: 380 passed; the only 3 failures are the same pre-existing, environment-specific ones as before (gitconfig quirk, WSL-vs-Windows path, `bat` availability).


jesperpedersen:
@tejassinghbhati I think the important is that

```
auto review immediate && export auto review
```

will work... so we have to look at having commands run without human interaction

tejassinghbhati:
@jesperpedersen Done. `auto review immediate && export auto review` works unattended now:

- `&&` chains commands in a schedule line, shell style. Each part runs in order, and if one fails the rest of its chain is dropped (the output window notes how many follow-ups were skipped).
- Scheduled commands run without human interaction. When the scheduler launches an auto review it starts at once and returns as soon as the run finishes, skipping the pre-start screen and the report browse that normally wait for a keypress. The report still lands in the output window and the clipboard, and the chained `export auto review` picks it up right after.
- The pending queue now tracks which commands came from the scheduler (they show a `(scheduled)` marker in `/pending`), which is also what tells auto review to run headless.

Also rebased onto main to pick up the Rust 1.97 changes and deep mode from #176. CI is green.

jesperpedersen:
Merged.

Issue #173, opened 2026-07-09
orangu-gguf


jesperpedersen:
Create an orangu-gguf binary.

The binary will detect CPU and GPU(s) on the machine together with memory and VRAM.

Then based on a HuggingFace URL it will suggest a command prompt for starting that model with `llama-server`.

It should should suggest the maximum model for the machine.

It should understand the roles: "all", "code", "review", "explorer", "embeddings" - we can suggest models upon startup after the CPU/GPU detect

jesperpedersen:
https://github.com/ggml-org/ggml/blob/master/docs/gguf.md

jesperpedersen:
Looking for people to help with this massive task

PR #177, opened 2026-07-10
[#173] orangu-gguf: Windows fixes for download symlinks and GPU listing

tejassinghbhati:
Part of #173. Two Windows-specific bugs found while testing the initial patch on a hybrid Intel/NVIDIA laptop (i5-12450H, Intel UHD + RTX 2050), plus one cosmetic fix:

- **`download`'s snapshot symlinks were dead on Windows.** The relative target was written with forward slashes (`../../blobs/<oid>`), which Windows cannot resolve: every native read through the link fails with "the filename, directory name, or volume label syntax is incorrect", so the model `orangu-gguf` itself had just fetched was invisible to `list`/`show`/the wizard and would be unreadable by llama-server. POSIX-emulating shells (Git Bash) resolve such links fine, which hides the problem. The target now uses backslashes on Windows. Verified: after the fix, `list` finds the downloaded model and `show`/the wizard read it.

- **`system` listed every NVIDIA card twice**: once from nvidia-smi and again from `Win32_VideoController`. The WMI pass now skips NVIDIA adapters when nvidia-smi already reported them (with better data: real VRAM use, and totals beyond `AdapterRAM`'s 32-bit cap), keeping them when nvidia-smi isn't available and WMI is the only source, mirroring the Linux sysfs scan's dedupe.

- Cosmetic: the report printed `[0] NVIDIA NVIDIA GeForce RTX 2050`; the vendor prefix is now skipped when the device name already leads with it.

All 64 orangu-gguf unit tests pass; clippy clean.

### DIFF
--- a/src/bin/orangu-gguf/download.rs
+++ b/src/bin/orangu-gguf/download.rs
@@ -592,6 +592,14 @@ fn link_or_copy(blob: &Path, link: &Path, oid: &str, file_path: &str) -> Result<
     // then two more reach the repo root, then descend into `blobs/<oid>`.
     let ups = "../".repeat(file_path.matches('/').count() + 2);
     let target = format!("{ups}blobs/{oid}");
+    // Windows only resolves symlink targets with backslash separators: a
+    // forward-slash target creates a link that Windows itself cannot follow
+    // (every native read fails with "the filename, directory name, or volume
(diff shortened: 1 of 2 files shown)


jesperpedersen:
Merged.

PR #180, opened 2026-07-10
[#173] Add dynamic model suggestion based on hardware limits

sksingh2005:


jesperpedersen:
@sksingh2005 See "orangu-gguf suggest"

jesperpedersen:
@sksingh2005 I think we will close this one

PR #182, opened 2026-07-11
[#173] orangu-gguf suggest: rename the Shared table to Combined

tejassinghbhati:
Part of #173, as discussed by mail.

- The second suggestion table is renamed from `Suggested model size (Shared)` to `Suggested model size (Combined)`: it sums every eligible GPU's total, dedicated and shared alike, so on a machine with both (my laptop: 4 GiB RTX 2050 + shared Intel UHD) it prints a budget larger than system RAM, which under a "(Shared)" heading reads like a bug at first glance.
- The docs (GGUF.md, manual chapters 45 and 77) follow the rename and now note the combined figure is inherently optimistic: the shared part of the pool is the same RAM the OS and everything else on the machine live in, so it's a hardware ceiling rather than a promise.
- `estimate_hidden_dims`'s documentation (and the matching passage in chapter 77) now says what the function really does: the params = 12 x layers x hidden^2 approximation is one equation with two unknowns, resolved by putting everything into the hidden size, so `layers` always works out to exactly 1 by construction; the KV estimate built on it scales as context x sqrt(params), which tracks GQA-era models well. The old comment read like it computed a real layer count.

No behavioral changes beyond the printed label. All 81 orangu-gguf tests pass (labels updated); clippy clean; verified the live output on my machine shows `(Combined)` with the same numbers as before.

### DIFF
--- a/doc/GGUF.md
+++ b/doc/GGUF.md
@@ -243,7 +243,7 @@ Suggested model size (Dedicated)
   128K     -                  -                    -
   256K     -                  -                    -
 
-Suggested model size (Shared)
+Suggested model size (Combined)
   Estimated budget : 66.17 GiB
 
   Context  Suggestion (Q2_K)  Suggestion (Q4_K_M)  Suggestion (Q8_0)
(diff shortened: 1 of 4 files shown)


jesperpedersen:
Merged.

PR #183, opened 2026-07-11
[#172] /build: Select a build target

### LINKED ISSUE #172: /build: Select target
Find a way to better select a build target - for example a specific `Makefile` rule

tejassinghbhati:
Closes #172

`/build` now takes an optional build target after the profile, in either order: `/build docs`, `/build release orangu-gguf`, `/build install debug`. Two profiles or two targets are rejected rather than one being silently dropped.

Each backend maps the target to its own native concept:

- **cargo**: a binary name, `--bin <target>` on both the build and its tests
- **CMake / Autotools**: a rule in the (generated) Makefile, `make <target>`
- **Meson**: `meson compile <target>`
- **Maven**: a lifecycle phase or goal, replacing the default `package` (`/build verify` runs `mvn verify`)
- **Go**: a package path, replacing the default `./...`
- **Python**: rejected with a clear error; `pip install -e .` has no target concept

A plain `Makefile` (checked as `GNUmakefile`, `makefile`, `Makefile` in GNU make's own lookup order) is now a last-resort backend of its own, so make-driven projects without a managed build system get `/build <rule>` too. It stays behind every managed backend in the detection chain since CMake, Autotools, and Meson all generate or ship Makefiles.

Tab completion (and the inline ghost hint) after `/build ` offers `debug`/`release` plus the workspace's discovered targets: cargo binary names (`[[bin]]` entries plus `src/bin/`) or Makefile rule names (skipping recipes, variable assignments, pattern rules, and the `.PHONY`-style dot rules). Only the token under the cursor completes, and discovery is best-effort so unlisted targets can still be typed.

Manual chapter and `/help` updated. Tests cover the request parser (order independence, duplicate rejection), the Makefile rule scanner, the lookup order, cargo/Makefile target discovery, completion behavior, and the existing command-parsing suite reworked for the new request type.


jesperpedersen:
Rebased and merged.

PR #186, opened 2026-07-12
[#171] /build: Incremental builds

### LINKED ISSUE #171: /build: Incremental builds
Improve support for incremental building of code - e.g. don't start over if the build environment hasn't changed

tejassinghbhati:
Closes #171

The generate-then-make backends (CMake, Autotools, Meson) no longer regenerate the build environment on every run. A repeat `/build` in the same profile with the same source-file set skips the reconfigure and reuses the existing object files instead of starting from scratch. Autotools was the worst case before, running `make distclean` and reconfiguring every single time.

The environment is regenerated only when it could actually be stale: the profile changed, or a source file was added or removed. That last case is the one a plain incremental `make` cannot recover on its own, since the generated build files still describe the old file set. Editing a file that already exists is left to the toolchain's own incremental rebuild, which already handles it.

To decide, each configure records the profile and the set of source files (tracked and untracked-but-not-ignored, via `git ls-files`) under orangu's per-workspace cache (`~/.orangu/workspace/<hash>/`), never in the source tree. When it cannot be sure the environment still matches (not a git repo, nothing recorded yet, or the file set differs) it regenerates, so a possibly-stale environment is never reused, which is the conservative default we agreed on in the issue.

- CMake: reconfigure when the cached `CMAKE_BUILD_TYPE` mismatches or the file set changed, else straight to `make`.
- Meson: `meson setup --reconfigure` only on a profile or file-set change, else straight to `meson compile` (previously it re-ran `meson configure` every build).
- Autotools: reuse an existing configuration with the same profile and file set, else `make distclean` + `configure`.
- Cargo and Go keep their own incremental caches and are unaffected.

Adds `BuildProfile::keyword` and the fingerprint helpers, with tests for the encode/decode round-trip, the reuse decision (profile change and file add/remove), and the git file listing (excludes gitignored build output). Manual updated.


jesperpedersen:
Rebased and merged.

PR #193, opened 2026-07-16
[#191] Surface session save errors in-app instead of silencing them

### LINKED ISSUE #191: Session chat history silently lost on workspace tab switch — errors discarded with let _
### Description

While reviewing the save-on-tab-switch code path, I noticed that
`save_session_messages()` and `update_session_metadata_branch()` at
`main.rs:1323-1327` and `main.rs:1368-1372` discard their `Result<()>` return
values with `let _ =`.

Currently, every other call site (e.g. `save_all_tabs!` at `main.rs:719`, and
`workspace_tab.rs:285`) correctly uses `?` to propagate errors.

### Impact

If the disk is full, the session directory is unwritable, or a write fails for
any reason, the user's entire session history is silently lost with zero
notification. The session *does* still exist in memory, but on tab switch or
workspace change the unsaved data is never persisted and the user has no way
to know.

### Root Cause

`save_session_messages()` (`session_store.rs:328`) calls `std::fs::write()`
which returns `io::Result`. `update_session_metadata_branch()`
(`session_store.rs:461`) calls `save_session_metadata()` which can also fail.
Both return `Result<()>`, but the callers at `main.rs:1323-1327` and
`main.rs:1368-1372` silence the error with `let _ =`.

### Proposed Approach

The fix should mirror the existing error-handling pattern used by
`WorkspaceTab::open()` at `main.rs:1350-1356` in the same function:

1. Replace `let _ = save_session_messages(...)` with
   `if let Err(err) = save_session_messages(...) {
       output_state.push_text(&format!("Error: {err:#}"));
   }`
2. Same for `update_session_metadata_branch(...)`
3. Apply to both the `OpenWorkspaceTab` handler (lines 1323-1327) and the
   `ChangeWorkspace` handler (lines 1368-1372)
4. Keep the `save_all_tabs!()` macro (used on `Quit`/`Restart` at line 719)
   unchanged — `?` is correct there since we're exiting anyway

The session remains alive in memory after the error, so the user can retry the
operation or `/copy` their work before switching.

I have a local patch ready if maintainers agree this is the correct direction.

AdeshDeshmukh:
Closes #191

Noticed this while tracing through the tab-switch flow in main.rs — save_session_messages() and update_session_metadata_branch() both return Result<()> but the callers in OpenWorkspaceTab and ChangeWorkspace just throw the value away with let _ =. If the disk is full or the session directory is unwritable, the user loses their entire conversation history with zero indication.

Everywhere else handles this right — save_all_tabs! on quit uses ?, workspace_tab.rs:285 uses ?. These two spots were just overlooked.

Changed them both to surface the error in the TUI via push_text(), same way WorkspaceTab::open() already handles its own failures a few lines below. If saving fails the user sees it and can retry or /copy their work before switching — nothing is lost since the session is still in memory.

save_all_tabs! stays as-is with ? since we're exiting anyway.

Also added myself to AUTHORS.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,3 +5,4 @@ Mahmoud Hamdy (TutTrue) <mahmoud.hamdy5113@gmail.com>
 Tejas Singh Bhati <tejassinghbhati077@gmail.com>
 Shashank Singh <shashanksgh3@gmail.com>
 Nishant raj Jha <nishantrajx924@gmail.com>
+Adesh Deshmukh <adeshkd123@gmail.com>
--- a/doc/manual/en/97-acknowledgement.md
+++ b/doc/manual/en/97-acknowledgement.md
@@ -12,6 +12,7 @@ Mahmoud Hamdy (TutTrue) <mahmoud.hamdy5113@gmail.com>
 Tejas Singh Bhati <tejassinghbhati077@gmail.com>
 Shashank Singh <shashanksgh3@gmail.com>
 Nishant raj Jha <nishantrajx924@gmail.com>
+Adesh Deshmukh <adeshkd123@gmail.com>
 ```
 
 ## Committers
--- a/src/bin/orangu/main.rs
+++ b/src/bin/orangu/main.rs
@@ -1320,11 +1320,24 @@ async fn run() -> Result<()> {
                     pending_tab_action = Some(TabAction::SwitchTo(index));
                     continue;
                 }
-                let _ = save_session_messages(&session_messages_path, session.messages());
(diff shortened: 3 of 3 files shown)


jesperpedersen:
@AdeshDeshmukh Remember doc/manual/en/97-acknowledgement.md

jesperpedersen:
@AdeshDeshmukh CI

jesperpedersen:
Merged.

jesperpedersen:
Welcome to the community !

PR #194, opened 2026-07-16
[#192] Propagate embedding cache write errors instead of silencing them

### LINKED ISSUE #192: Embedding index rewrite_cache silently discards all write failures — cache corruption goes undetected
### Description

While reviewing the embedding cache persistence code path, I noticed that
`rewrite_cache()`, `write_meta()`, `append_chunks()`, and `append_manifest()`
in `src/embeddings/index.rs` silently discard every I/O error with `let _ =`.

Every call to `std::fs::create_dir_all()` and `std::fs::write()` inside these
functions — 8 `let _` instances across lines 426, 434, 446, 447, 664, 678,
700, and 707 — can fail without any error reaching the user or the caller.

### Impact

If the disk is full, the cache directory is unwritable, or a write fails for
any reason, the embedding index is silently corrupted or goes missing. The
user must re-index from scratch with no explanation. For large workspaces a
full re-index takes minutes.

### Root Cause

| Function | Line(s) | Silenced call |
|----------|---------|---------------|
| `rewrite_cache` | 426 | `std::fs::create_dir_all(dir)` |
| `rewrite_cache` | 434 | `std::fs::write(chunks_path, body)` |
| `write_meta` | 446 | `std::fs::create_dir_all(dir)` |
| `write_meta` | 447 | `std::fs::write(meta_path, json)` |
| `append_chunks` | 664 | `std::fs::create_dir_all(parent)` |
| `append_chunks` | 678 | `f.write_all(body.as_bytes())` |
| `append_manifest` | 700 | `std::fs::create_dir_all(parent)` |
| `append_manifest` | 707 | `f.write_all(body.as_bytes())` |

All four functions currently return `()`. The enclosing function
`build_or_update()` (line 217) already returns `Result<Self>`, so propagating
errors upward is natural.

### Proposed Approach

The fix mirrors the existing `Result<()>` return pattern used by every other
fallible function in the file:

1. Change `rewrite_cache(&self, dir: &Path)` → return `Result<()>`, replace
   `let _ =` with `?` at lines 426 and 434
2. Change `write_meta(&self, dir: &Path)` → return `Result<()>`, replace
   `let _ =` with `?` at lines 446 and 447
3. Change `append_chunks(path, chunks)` → return `Result<()>`, replace
   `let _ =` with `?` at lines 664 and 678
4. `append_manifest` kept as best-effort with `warn!` logging — its existing
   documentation states `"Errors are ignored; the manifest is a progress aid,
   never load-bearing"`, so aborting the entire index build over a log write
   is disproportionate
5. Add `?` at each call site inside `build_or_update()`: lines 324, 378, 417

The `if let Ok` silent skips (serialization failures at lines 429, 445, 668;
file-open failures at lines 673, 702) are a related concern but out of scope
for this fix.

I have a local patch ready and would be glad to adjust it to whatever strategy
the project maintainers prefer.

AdeshDeshmukh:
Closes #192

Noticed this while tracing through the embedding index persistence code in index.rs — rewrite_cache(), write_meta(), and append_chunks() all throw away their I/O errors with let _ =. If the disk is full or the cache directory is unwritable, the embedding index silently goes missing and the user has to re-index from scratch with no explanation.

Changed all three to return Result<()> and propagate errors up with ?. The call sites in build_or_update() are updated to match.

append_manifest() is kept best-effort — its docs already say it is a progress aid, never load-bearing. Aborting the entire embedding pipeline over a log write would be the wrong tradeoff.

### DIFF
--- a/src/embeddings/index.rs
+++ b/src/embeddings/index.rs
@@ -321,7 +321,7 @@ impl EmbeddingIndex {
             module_sketches: HashMap::new(),
             chunks: kept,
         };
-        index.rewrite_cache(&dir);
+        index.rewrite_cache(&dir)?;
 
         // ── Phase 2 (50–100%): upload — embed the parsed chunks concurrently ──
         // The upload to the embedding server is the bottleneck, so keep up to
@@ -375,7 +375,7 @@ impl EmbeddingIndex {
                     let Some(joined) = maybe else { break };
                     let (_rel, _hash, embedded, bytes) = joined??;
                     // The file's hash is already in meta; append its chunks.
-                    append_chunks(&chunks_path, &embedded);
+                    append_chunks(&chunks_path, &embedded)?;
                     index.chunks.extend(embedded);
                     done_bytes = done_bytes.saturating_add(bytes);
                     let permille = (500 + done_bytes.saturating_mul(500) / total_bytes).min(1000);
@@ -414,7 +414,7 @@ impl EmbeddingIndex {
                 .module_sketches
                 .insert(file, super::sketch::EllipsoidSketch::compute(&vectors));
         }
-        index.write_meta(&dir);
+        index.write_meta(&dir)?;
 
         Ok(index)
     }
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Rebased and merged.

Issue #198, opened 2026-07-17
git_commit_count silently returns Ok(0) on parse failure, bypassing push-behind guard


AdeshDeshmukh:
### Description

While reviewing the push-behind guard code path, I noticed that
`git_commit_count()` at `forge.rs:838-851` parses `git rev-list --count`
output with `.parse().unwrap_or(0)`. If the command produces unexpected
output — an error message on stdout, a malformed number, or a locale issue
— the function silently returns `Ok(0)` instead of `Err(...)`.

### Impact

The caller `git_push()` at `ops.rs:823-831` uses this to check whether the
branch is behind its base ref. A return of `0` means "not behind," so the
stale-branch guard is silently bypassed. Users can push stale branches
without the required rebase warning, potentially overwriting remote history
or getting rejected with a confusing error from the remote.

### Root Cause

The function checks neither `output.status.success()` nor handles parse
failure properly:

```rust
pub(crate) fn git_commit_count(repo_root: &Path, range: &str) -> Result<usize> {
    Ok(String::from_utf8_lossy(
        &std::process::Command::new("git")
            .arg("-C")
            .arg(repo_root)
            .args(["rev-list", "--count", range])
            .output()
            .context("failed to run git rev-list")?
            .stdout,
    )
    .trim()
    .parse()
    .unwrap_or(0))
}
```

Every other `git_*` function in the file returns `Result` with proper
error context. The function already declares `Result<usize>` as its return
type, so returning `Err` is the existing contract.

### Proposed Approach

1. Check `output.status.success()` before parsing — return `Err` with
   stderr context if git exited non-zero
2. Replace `.unwrap_or(0)` with `.context()` so parse failures propagate
   as `Err` instead of silently returning `0`
3. Include the range in error messages for debuggability

I have a local patch ready and would be glad to adjust it to whatever
strategy the project maintainers prefer.

jesperpedersen:
@AdeshDeshmukh Go ahead by submitting a pull request

PR #202, opened 2026-07-18
[#198] propagate git_commit_count parse failures instead of returning 0

AdeshDeshmukh:
Propagate `git rev-list --count` failures (non-zero exit or unparseable stdout) as errors instead of silently returning 0.\n\nCloses #198


jesperpedersen:
Rebased and merged.

PR #203, opened 2026-07-18
[#199] reject git_squash when working tree is dirty

### LINKED ISSUE #199: git_squash silently excludes uncommitted changes — no dirty-tree check before destructive reset
### Description

While reviewing the `/squash` command path, I noticed that `git_squash()` at
`ops.rs:918-1028` runs `git reset --soft <merge_base>` followed by
`git commit -m` with zero validation of the working tree state. If the
working tree has uncommitted changes, `reset --soft` moves HEAD but leaves
the index as-is — the subsequent commit captures only the last committed
state, not the dirty changes. The uncommitted work remains as unstaged
modifications and is easy to lose.

### Impact

Users can lose uncommitted work when running `/squash`. There is no warning
or error — the squash appears to succeed, but the dirty changes are silently
dropped from the resulting commit.

### Root Cause

The function checks whether the branch is protected and whether there are
commits to squash, but never inspects the working tree for uncommitted
changes before performing the destructive `reset --soft`.

### Proposed Approach

Add a `git status --porcelain` check at the start of `git_squash()`, right
after the protected branch and merge-base checks (line 924). If the output
is non-empty, return an error listing the dirty files and asking the user
to commit or stash first.

This is the standard pattern for dirty-tree validation before destructive
git operations across the Rust ecosystem.

I have a local patch ready if maintainers agree this is the correct direction.

AdeshDeshmukh:
Add a dirty-tree check at the start of `git_squash()` to reject the operation when the working tree has uncommitted changes.

`git reset --soft` followed by `git commit -m` silently drops uncommitted work — the squash appears to succeed but the dirty changes are excluded from the resulting commit.

The existing `count_pending_changes()` helper (used by `pending_changes_summary()` and `status_output()`) already runs `git status --porcelain`. This change calls it before any destructive operation and returns a clear error if the tree is dirty.

Closes #199

## Testing Performed
- [x] `cargo build` — compiles
- [x] `cargo clippy` — no warnings
- [x] `cargo test -p orangu --lib` — all 277 passed

### DIFF
--- a/src/bin/orangu/git/ops.rs
+++ b/src/bin/orangu/git/ops.rs
@@ -923,6 +923,14 @@ pub fn git_squash(repo_root: &Path) -> Result<String> {
 
     let base_ref = git_find_base_ref(repo_root)?;
 
+    let (dirty_count, _) = count_pending_changes(repo_root)?;
+    if dirty_count > 0 {
+        return Err(anyhow!(
+            "working tree has {dirty_count} uncommitted change(s); \
+             commit or stash them before squashing"
+        ));
+    }
+
     let merge_base_output = std::process::Command::new("git")
         .arg("-C")
         .arg(repo_root)


jesperpedersen:
Rebased and merged.

PR #207, opened 2026-07-18
[#206] validate model name against available_models in SetModelId

### LINKED ISSUE #206: SetModelId silently accepts any model name — validation against available_models missing
### Description

`/model <name>` at `dispatch.rs:413-421` silently accepts any model name without checking if it exists in the list of models the server advertises.

### Impact

If a user types `/model unicorn-v99`, the model ID is set without validation. Inference later fails with a confusing server error (or silently returns bad output). The user had no indication the model name was invalid.

### Root Cause

The handler only checks if the name is empty:
```
if name.is_empty() {
    return Ok(CommandOutcome::OutputError(...));
}
*active_model_id = name.to_string();
```

Compare with `SetServer` at line 469 which validates against the `llms` map:
```
if !llms.contains_key(name) {
    return Ok(CommandOutcome::OutputError(
        "Unknown server '{name}'. Available: ..."
    ));
}
```

### Proposed Fix

Add a validation check against `available_models` in the `SetModelId` handler, following the same pattern as `SetServer`. Skip validation when `available_models` is empty (coordinator not yet resolved).

### Files That Need Changing

- `src/bin/orangu/dispatch.rs:413-421` — ~5 lines added

### Out of Scope

- Adding `session.set_system_prompt()` — not needed, server doesn't change
- Adding `*detect_model = true` — not needed, capabilities don't change
- These are intentionally different from `SetServer` behavior

AdeshDeshmukh:
Add validation to /model (SetModelId) so that setting a non-existent model ID returns a clear error instead of silently failing at inference time.

**Before:** /model unicorn-v99 silently sets the ID — inference fails later with a confusing server error.
**After:** /model unicorn-v99 immediately returns Unknown model 'unicorn-v99'. Available: model-a, model-b

Follows the same validation pattern as SetServer. Skips validation when available_models is empty (server doesn't advertise models / coordinator not yet resolved).

## Testing
- [x] cargo build — compiles
- [x] cargo clippy — no warnings
- [x] cargo test -p orangu --bin orangu — 414 passed (1 new test)
- [x] Added set_model_rejects_unknown_model_name test
- [x] Manual: /model unicorn-v99 -> Unknown model 'unicorn-v99'

## AI Disclosure
This PR was assisted by AI. I have reviewed and tested every line.

Closes #206

### DIFF
--- a/src/bin/orangu/dispatch.rs
+++ b/src/bin/orangu/dispatch.rs
@@ -416,6 +416,12 @@ pub(crate) fn handle_command(
                     model_usage_message().to_string(),
                 ));
             }
+            if !available_models.is_empty() && !available_models.iter().any(|m| m == name) {
+                return Ok(CommandOutcome::OutputError(format!(
+                    "Unknown model '{name}'. Available: {}",
+                    available_models.join(", ")
+                )));
+            }
             *active_model_id = name.to_string();
             save_session_settings(session_dir, Some(active_model), Some(active_model_id));
             Ok(CommandOutcome::Quiet)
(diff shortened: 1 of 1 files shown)


jesperpedersen:
Remember to rebase your patches, it is becoming difficult to merge your code

jesperpedersen:
Rebased and merged.

jesperpedersen:
So, it takes longer for me to review and merge

PR #209, opened 2026-07-20
[#208] wrap /log and /fetch output in LLM context

### LINKED ISSUE #208: /log and /fetch commands skip LLM context — inconsistent with /diff and /show
Log (dispatch.rs:726) and Fetch (dispatch.rs:763) return CommandOutcome::Output, while Diff (line 507) and Show (line 730) wrap in OutputWithLlmContext using prepare_llm_diff_context_with_stats() + compression metrics.

The LLM never sees git log or git fetch output in its context.

Expected: Both should wrap their output the same way so the LLM sees git log and git fetch results in context.

AdeshDeshmukh:
## What
Wrap /log and /fetch output in OutputWithLlmContext so the LLM sees their results (matching /diff and /show).

## Why
Log and Fetch return bare CommandOutcome::Output. The LLM never sees git log or git fetch output.
Diff and Show use prepare_llm_diff_context_with_stats + OutputWithLlmContext — the LLM sees those.
This is a design inconsistency that silently loses useful context.

## How
Both handlers now:
1. Call prepare_llm_diff_context_with_stats() for compression (same as Diff/Show)
2. Record compression metrics via tools.compression_metrics
3. Build llm_context: "The user executed `/log{count?}`. Output:\n\n..."
4. Return OutputWithLlmContext { display: raw_output, llm_context }

Note: Using prepare_llm_diff_context_with_stats because (a) matches Diff/Show pattern,
(b) compress_git_diff gracefully passes non-diff text (log/fetch output) through unchanged,
(c) diff_file_cap only activates on actual diff content.

## Testing
- [x] cargo build — compiles
- [x] cargo clippy — no warnings
- [x] cargo test -p orangu --bin orangu — all pass
- [x] Added log_command_wraps_output_in_llm_context — success path with real git repo
- [x] Added log_without_git_repo_returns_error — error-path regression test
- [x] Manual: /log shows raw output, LLM context contains "The user executed `/log`. Output:..."

## Out of Scope
- Stripping ANSI color codes from git output (pre-existing, affects all commands)
- Creating a generic prepare_llm_context_with_stats function
- Adding LLM context for /bisect_log or other secondary git commands

## AI Disclosure
This PR was assisted by [tool]. I have reviewed and tested every line.

Closes #208

### DIFF
--- a/src/bin/orangu/dispatch.rs
+++ b/src/bin/orangu/dispatch.rs
@@ -724,7 +724,32 @@ pub(crate) fn handle_command(
             ))
         }
         LocalCommand::Log(count) => match log_output(workspace, count) {
-            Ok(output) => Ok(CommandOutcome::Output(output)),
+            Ok(output) => {
+                let (context, stats) =
+                    orangu::compression::prepare_llm_diff_context_with_stats(
+                        &output,
+                        compression,
+                        tools.diff_file_cap(),
+                        Some(&*tools.compression_store),
+                    );
+                if let Ok(mut metrics) = tools.compression_metrics.lock() {
+                    metrics.record(&stats);
+                }
+                let count_part = count.map(|c| format!(" {c}")).unwrap_or_default();
+                let mut llm_msg =
+                    format!("The user executed `/log{count_part}`. Output:\n\n");
+                if let Some(note) = context.note {
+                    llm_msg.push_str(&note);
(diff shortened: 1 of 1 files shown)


jesperpedersen:
@AdeshDeshmukh How is this going ?

jesperpedersen:
@AdeshDeshmukh Reopen, if you want

PR #215, opened 2026-07-25
Fix GPU memory type detection on Apple Silicon

### LINKED ISSUE #216: GPU memory type shows Unknown/n/a on Apple Silicon
orangu-server system reports Memory type: Unknown and VRAM total: n/a on Apple Silicon Macs (confirmed on an M5 Pro, macOS 26.5.2).

Cause: macos_memory_kind() in src/hardware.rs decides Dedicated vs Shared from system_profiler's spdisplays_vram and spdisplays_vram_shared keys. On this machine, system_profiler SPDisplaysDataType -json has neither key at all for the GPU entry, so it falls through to Unknown, and VRAM total stays empty since the Shared-to-system-RAM fallback never triggers.

Fix in progress: #215.

bassamadnan:
Fixes #216.

## Summary

- macos_memory_kind() reported Unknown/n/a on an M5 Pro because system_profiler's JSON output there has neither spdisplays_vram nor spdisplays_vram_shared.
- Every Apple Silicon Mac is unified memory, no discrete-VRAM model exists, so aarch64 now treats a missing key as Shared. x86_64 (Intel Mac) behavior is unchanged.
- Limited to Apple: the aarch64 check only runs inside macos_memory_kind(), which is itself gated to target_os = "macos", so this cannot affect any non-Apple aarch64 platform.

Before:
```
GPU
  [0] Apple M5 Pro
      Memory type  : Unknown
      VRAM total   : n/a
```

After:
```
GPU
  [0] Apple M5 Pro
      Memory type  : Shared
      VRAM total   : 24.00 GiB
```

## Test plan

- [x] orangu-server system shows Shared / 24.00 GiB on an M5 Pro
- [x] cargo test --lib hardware:: passes, including a new aarch64-only test and the existing x86_64 test (unchanged assertion)
- [x] cargo fmt --all --check and cargo clippy -- -D warnings both clean

### DIFF
--- a/src/hardware.rs
+++ b/src/hardware.rs
@@ -429,17 +429,17 @@ fn detect_macos_gpus() -> Vec<GpuInfo> {
         .collect()
 }
 
-/// `system_profiler`'s own two keys already say which kind of memory this
-/// is: `spdisplays_vram` names a real dedicated-VRAM figure (a discrete
-/// card, e.g. an eGPU or an older Mac Pro/MacBook Pro), while
-/// `spdisplays_vram_shared` marks Apple Silicon's unified-memory
-/// architecture or an older Intel Mac's integrated graphics — either way,
-/// memory shared with the CPU rather than a separate pool.
+/// `spdisplays_vram` means dedicated, `spdisplays_vram_shared` means
+/// unified/shared. Some Apple Silicon machines omit both keys (confirmed
+/// on an M5 Pro), so `aarch64` treats that as `Shared` too: every Apple
+/// Silicon Mac is unified memory, no discrete-VRAM model exists. Intel
+/// Macs (`x86_64`) keep the old `Unknown` fallback, since that
+/// architecture shipped both integrated and discrete GPUs.
 #[cfg(target_os = "macos")]
 fn macos_memory_kind(entry: &serde_json::Value) -> MemoryKind {
     if entry.get("spdisplays_vram").is_some() {
         MemoryKind::Dedicated
-    } else if entry.get("spdisplays_vram_shared").is_some() {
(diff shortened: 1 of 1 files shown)


src/hardware.rs
```diff
@@ -734,13 +734,24 @@ mod tests {
...
         assert_eq!(macos_memory_kind(&dedicated), MemoryKind::Dedicated);
 
         let shared = serde_json::json!({"spdisplays_vram_shared": "spdisplays_unified"});
         assert_eq!(macos_memory_kind(&shared), MemoryKind::Shared);
+    }
 
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    #[test]
+    fn macos_memory_kind_assumes_shared_on_apple_silicon_without_either_key() {
+        let neither = serde_json::json!({"_name": "Apple M5 Pro"});
```
jesperpedersen: What about the other Macs ?

bassamadnan: ahh i thought thats just for tests...checking

bassamadnan: Correct, that string is only test data (unused by the logic), now renamed to "Apple GPU" to make that clearer.

The fix itself checks target_arch = "aarch64", not the chip name, so it applies identically to every Apple Silicon Mac (M1 through M5, any future chip), not just the M5 Pro this was confirmed on. It's also safe either way:

- If a machine's system_profiler still reports spdisplays_vram_shared, that branch runs first and already returns Shared, the new aarch64 fallback never gets reached.
- If the key is missing (confirmed on this M5 Pro), the fallback catches it and still resolves to Shared.

Both paths converge on the same answer, so this can't regress any Apple Silicon Mac, tested or not.

jesperpedersen:
Merged.

PR #217, opened 2026-07-27
Refuse OpenCL backend on Apple targets instead of crashing

bassamadnan:
## Summary

- Apple's own OpenCL ICD reports a device, so OpenClBackend::try_init succeeds, but the first real matmul call segfaults inside clSetKernelArg (confirmed on an Apple Silicon Mac). This crashed backend = auto, an explicit backend = opencl, and cargo test itself (engine::backend::opencl::tests, which assumed try_init always returns None on a machine with no vendor ICD registered).
- try_init now refuses to initialize at all on Apple targets, so every caller (backend selection and the test suite) gets the same graceful None it already handles everywhere else. select_backend's Auto case drops the now-redundant Apple-specific check this replaces.
- Also gates vulkan.rs's test-only capture helpers (only consumed by vulkan_replay.rs's Vulkan-only tests, already excluded on Apple) to match, fixing a cargo clippy --all-targets dead-code failure on macOS.
- Scoped to target_vendor = "apple" only: every change either compiles to nothing on other platforms or, in main.rs, restores the original unconditional code. No other backend or platform is affected.

## Test plan

- [x] cargo test --bin orangu-server on an Apple Silicon Mac: was a SIGSEGV crash, now 349 passed, 0 failed, 0 crashes
- [x] cargo fmt --all --check clean
- [x] cargo clippy --all-targets -- -D warnings clean (was failing with dead-code errors on macOS)
- [x] orangu-server system and a real chat completion still work correctly with backend = auto

### DIFF
--- a/src/bin/orangu-server/engine/backend/opencl.rs
+++ b/src/bin/orangu-server/engine/backend/opencl.rs
@@ -23,7 +23,10 @@
 //! registered, so [`OpenClBackend::try_init`] finds zero platforms here and
 //! gracefully returns `None`, the same as every other machine with no
 //! OpenCL device — this module's cross-check tests skip in exactly that
-//! case, per the same convention `vulkan.rs`/`cuda.rs` use.
+//! case, per the same convention `vulkan.rs`/`cuda.rs` use. Apple targets
+//! are the one exception: their OpenCL ICD does report a device, but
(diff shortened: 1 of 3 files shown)


jesperpedersen:
Rebased and merged.

Issue #168, opened 2026-07-08
MCP integration


jesperpedersen:
We should provide support for basic integration with Model Context Protocol (MCP) servers in order to get their functionality

jesperpedersen:
I'll go through this in detail before closing

PR #224, opened 2026-08-01
[#168] Add local MCP management, refresh, and reconnect support

sksingh2005:


jesperpedersen:
@sksingh2005 The configuration is way too hard... Think 

```
[pgmoneta]
mcp = true
endpoint = http://localhost:8100
...
```

@ahmed-kamal2004 Can you go over this ?

jesperpedersen:
@sksingh2005 We will need web ui for this as well

ahmed-kamal2004:
> @sksingh2005 The configuration is way too hard... Think
> 
> ```
> [pgmoneta]
> mcp = true
> endpoint = http://localhost:8100
> ...
> ```
> 
> @ahmed-kamal2004 Can you go over this ?

I will review this PR tmw.

jesperpedersen:
@sksingh2005 I reserved

```
[orangu-server]
...

[web]
...
```

I think we can assume that the rest are MCP servers

sksingh2005:
> @sksingh2005 We will need web ui for this as well

@jesperpedersen I did not understood this one , should we let web-chat model calls use MCP tools or add a borswer ui to edit the local client's conf file ? 

Since the existing web ui neither reads conf file nor has a tool-call loop ig .

jesperpedersen:
@sksingh2005 I think adding/modifying/deleting a MCP from the server should require a restart - so it should just be a list of available MCPs, and a "Show" button if we can get more information -- see the "Show" button for the "Models"

src/mcp.rs
```diff
@@ -0,0 +1,501 @@
...
+struct McpState {
+    definitions: Vec<ToolDefinition>,
+    routes: HashMap<String, Route>,
+    status: Vec<String>,
+    instructions: String,
+}
+
+impl McpManager {
+    /// Launch every configured local server. A single failure is returned as a
+    /// warning instead of preventing the workspace from opening.
```
ahmed-kamal2004: I am thinking if a startup failure could be better to eliminate ambiguous state

src/bin/orangu/wait.rs
```diff
@@ -77,6 +78,18 @@ pub(crate) async fn wait_for_response(
                         state.native_tool_calls.push(tool_call.clone());
                     }
                 },
+                move |tool_call| {
+                    let (decision, receiver) = std::sync::mpsc::channel();
+                    if let Ok(mut state) = sta.lock() {
+                        state.pending_mcp_approval =
+                            Some(crate::input::McpApprovalRequest { decision });
+                        state.output.push_str(&format!(
+                            "\n\n[MCP approval required: {}. Press y to approve or n to deny.]",
```
ahmed-kamal2004, line 87: I am thinking about if having the `arguments` of the tool shown in the approval would be a good thing.

src/mcp.rs
ahmed-kamal2004: Since there is `required` option per MCP, this comment is outdated.

src/bin/orangu/dispatch.rs
```diff
@@ -520,6 +520,53 @@ pub(crate) fn handle_command(
...
+                    Err(err) => Ok(CommandOutcome::OutputError(format!(
+                        "MCP configuration error: {err:#}"
+                    ))),
+                }
+            }
+            McpSubcommand::Add {
+                name,
+                command,
+                args,
+            } => orangu::config::add_mcp_server(config_path, &name, &command, &args)
```
ahmed-kamal2004: `add_mcp_server` is only for `stdio` mcps ? not `http-sse` ones ? 
may I am missing something, what could be the reason for this.

ahmed-kamal2004 (commented):
The idea of lifecycle management of the `stdio` mcps per workspace and recreation on connection would be harmful for stateful mcps, but now since there is a new release few days ago focusing on stateless mcps instead of stateful ones, then we can say it is safe now 
https://aaif.io/blog/mcp-is-growing-up#b15

ahmed-kamal2004:
@sksingh2005 I see great work is done here, Well done!

@jesperpedersen I started thinking about a new binary responsible for mcp aggregation/adaptor

Description in one line: Unify multiple MCPs (`stdio` - `http`) into one `http` MCP

Details: it will be responsible for the lifecycle of the `stdio` mcps, handling connections and reconnections, authentication...etc

I see this is helpful if we going f.ex to run many MCPs on a production machine/cluster
instead of managing each alone, we separate concerns/responsibilities.

jesperpedersen:
@sksingh2005 We just need to support running MCP servers - not being able to start them

jesperpedersen:
@sksingh2005 At run-time we can add and modify - not remove

doc/etc/orangu.conf
```diff
@@ -41,3 +41,24 @@ model = unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF
...
 # endpoint = http://localhost:8300/v1
 # model = ggml-org/embeddinggemma-300M-GGUF
+
+# When [orangu-server] or [web] is present, every other section is an HTTP MCP
+# service. `endpoint` is the server's MCP URL, not its general HTTP port.
+# [pgmoneta]
+# endpoint = http://localhost:8100/mcp
+# approval_mode = writes
+
+# Advanced local stdio Model Context Protocol tools. `command` is launched directly in
```
jesperpedersen: We can remove all of this

doc/CONFIGURATION.md
```diff
@@ -80,3 +80,56 @@ Use `/server` to switch between the configured servers at runtime; Tab
...
+## MCP servers
+
+When `[orangu-server]` or `[web]` is present, they are the two LLM profiles;
+every other section is an HTTP MCP service. No flag, command, or argument
+quoting is needed:
+
+```ini
+[pgmoneta]
+endpoint = http://localhost:8100/mcp
+approval_mode = writes
```
jesperpedersen: What is this ?

jesperpedersen:
@sksingh2005 You need a rebase, and cleanup.

We need to make this as simple as possible

doc/etc/orangu.conf
```diff
@@ -24,13 +24,13 @@ theme = classic         # classic|modern_dark|modern_light|oranguday|tokyonight|
...
 # Each server must use a unique endpoint.
-[main-server]
+[orangu-server]
 role = all
 endpoint = http://localhost:8100/v1
 model = ggml-org/gemma-4-E4B-it-GGUF
 # api_key = sk-...    # required if the server runs with --api-key
 
-[qwen-server]
+[web]
```
jesperpedersen: This isn't correct

jesperpedersen:
@sksingh2005 Maybe just do `-i` and copy the config in

jesperpedersen:
Rebased and merged.

Issue #238, opened 2026-08-12
Web console: Save as


jesperpedersen:
The web console should have its "Save" button as

* Markdown (Default)
* Text
* PDF

PR #239, opened 2026-08-12
[#238] Add answer save formats

sksingh2005:


jesperpedersen:
@sksingh2005 It looks good - but the dropdown is sort of transparent over the existing text. "Markdown (default)" -> "Markdown". We should open the PDF in a new window

jesperpedersen:
Rebased and merged.

Issue #264, opened 2026-09-14
Prometheus support


jesperpedersen:
Add Prometheus support to orangu-server.

Use `metrics` to enable it

Saleh-enab:
Hi, I’d like to work on this feature

My understanding of the current implementation is that the project already has a custom `ServerMetrics` struct and a manually implemented `/metrics` endpoint. My plan would be to introduce the standard Rust `metrics` facade together with `metrics-exporter-prometheus`, install the Prometheus recorder at startup, and use the existing `/metrics` endpoint to expose the metrics

One thing I’d like to ask about before starting which is, should the existing `ServerMetrics` be fully replaced by the new metrics system, or should it be kept alongside the new instrumentation?

jesperpedersen:
@Saleh-enab Go for it !

jesperpedersen:
@Saleh-enab Think of this as a separate port in orangu-server.conf to expose the Prometheus formatted data,

```
...
metrics = 8300
...
```

The current `/metrics` is for the current stack - just that somebody want to collect Prometheus data from all the orangu-servers on the network.

We can likely share data structures... Remember documentation, and `-i / --init` support. Default is off

PR #271, opened 2026-09-16
[#264] orangu-server: Add dedicated Prometheus metrics port

Saleh-enab:
## Summary

Adds an optional, unauthenticated Prometheus `/metrics` listener on its own port — off by default, and separate from the existing key-protected `/metrics` on the API port.

- New `[orangu-server].metrics` config key (default `0` = disabled) and matching `--metrics-port` CLI flag.
- Reuses the existing Prometheus rendering code (`render_metrics_text`) so both endpoints serve identical text — no new metric format, only a new listener.
- The dedicated listener is bound and spawned the same way the optional web console listener already is, and deliberately carries no `require_api_key` layer — that's the whole point of a second port.
- Survives a web-console-triggered model reload (reexec) by inheriting the socket across the restart, same as the API and web listeners already do.
- `bundle` executables can bake in a default metrics port the same way they already do for host/port/web.
- `-i`/`--init` gained a new yes/no prompt to enable it; declining writes nothing, so existing config files are unaffected.
- Startup banner prints a `Metrics` line (`disabled`, or the bound address with a no-API-key note).
- Updated bash/zsh/fish completions, the manual, `SERVER.md`, the example config, and the local setup guide.


jesperpedersen:
@Saleh-enab You need the authors files... see DEVELOPERS.md

Saleh-enab:
@jesperpedersen sorry for missing this part. I’ve updated it now

jesperpedersen:
@Saleh-enab Look at how `[web]` is done and do the same for `[prometheus]`

jesperpedersen:
@Saleh-enab Make sure that the latest Prometheus version can scrape the endpoint without errors

jesperpedersen:
@Saleh-enab Add simple smoke tests for that

Saleh-enab:
> @Saleh-enab Look at how `[web]` is done and do the same for `[prometheus]`

 @jesperpedersen  Do you mean having a separate `[prometheus]` section in the config file, similar to `[web]`, instead of just adding it as a field?

I’ve implemented it following the same configuration structure and rules as the web console, so it should already follow the same approach. If you have something specific in mind that should be different, could you clarify it please?

jesperpedersen:
@Saleh-enab Yes, do it as https://github.com/mnemosyne-systems/orangu/blob/main/doc/SERVER.md#the-web-section - this is the web console

Saleh-enab:
@jesperpedersen I’ve re-implemented it in the same way as `[web]`, as you suggested, and tested it against the latest Prometheus version, `v3.14.0`. Everything works without any issues.

Regarding the smoke test you asked for, I don’t think there’s a way to run a meaningful test without involving both the `orangu-server` binary and the Prometheus binary to verify that everything works correctly.

Is that what you had in mind for the smoke test?

jesperpedersen:
@Saleh-enab You need to rebase - keep an eye on the repository because a lot of patches are coming in

Saleh-enab:
@jesperpedersen Done

jesperpedersen:
orangu-server -i

```
Add a dedicated Prometheus metrics endpoint (separate port, blank = false by default) [y/N]: 
```

it should follow the web console style

jesperpedersen:
There is no `host` question

jesperpedersen:
```
Metrics    http://0.0.0.0:8300/metrics (no API key)
```

There is no API key for this

jesperpedersen:
@Saleh-enab See if you can make a simple embedded front-page (`...:8300/`) so we don't get 404 on that one

Saleh-enab:
@jesperpedersen 
Done, I’ve updated the metrics port question to follow the exact same style as the web console. I also removed the text mentioning “no API key” to avoid confusing users, since this port doesn’t require any API keys as you said. Additionally, I added a simple HTML page to serve the metrics port’s index route, as you requested.

src/bin/orangu-server/shell.rs
```diff
@@ -110,14 +110,14 @@ _orangu_server() {
             COMPREPLY=( $(compgen -W "off auto all cpu" -- "$cur") )
             return 0
             ;;
-        -p|--port|--web|--device|--threads)
+        -p|--port|--web|--device|--threads|--metrics-port)
```
jesperpedersen: Just call it `--metrics`

jesperpedersen:
@Saleh-enab Just follow `--web`

Saleh-enab:
@jesperpedersen Done, I've updated it to be just `--metrics`

Saleh-enab:
@jesperpedersen I’ve also added a custom `host` field, similar to the one already available in the `[web]` section. As requested, the `[prometheus]` section now follows the `[web]` section

jesperpedersen:
@Saleh-enab I'm busy with other patches, so keep rebasing. And, look at what metrics we can add like

* Project name
* Project version
* Uptime
* ...

jesperpedersen:
@Saleh-enab You can likely get ideas from a pgexporter setup

Saleh-enab:
I'm on it

Saleh-enab:
@jesperpedersen I've add some new process related metrics as you suggested which are:

- `orangu_server_build_info` — Build/version information for the running Orangu server.
- `orangu_server_host_info` — OS, kernel, and architecture information of the host.
- `orangu_server_uptime_seconds` — Time elapsed since the server started.
- `process_start_time_seconds` — Unix timestamp when the server process started.
- `process_cpu_seconds_total` — Total CPU time consumed by the server process.
- `process_resident_memory_bytes` — Physical memory currently used by the server process.
- `process_virtual_memory_bytes` — Virtual memory currently allocated to the server process.
- `orangu_server_cpu_logical_cores` — Number of logical CPU cores available to the server.
- `orangu_server_host_memory_total_bytes` — Total physical memory available on the host.
- `orangu_server_host_memory_available_bytes` — Physical memory currently available on the host.
- `orangu_server_http_requests_in_flight` — Number of HTTP requests currently being handled.
- `orangu_server_slots_idle` — Number of request slots currently available.

jesperpedersen:
@Saleh-enab Excellent !

* Maybe fold `orangu_server_build_info` and `orangu_server_host_info` into `orangu_server_host_info`
* Fold `process_start_time_seconds` into `orangu_server_uptime_seconds` and rename to `orangu_server_uptime`
* `process_cpu_seconds_total` -  have it include the load numbers, rename to `orangu_server_cpu`
* Maybe `orangu_server_http_requests_in_flight` to `orangu_server_http_requests` and have all the statics there

Saleh-enab:
@jesperpedersen Done

Saleh-enab:
@jesperpedersen The linting check is currently failing due to a Clippy warning in "src/bin/orangu-server/engine/kv_pool.rs":

"error: this loop could be written as a while let loop"

This appears to be related to code that was previously edited in "kv_pool.rs". I did not modify this file as part of my changes, so I believe the issue may have been introduced by an earlier edit

Saleh-enab:
@jesperpedersen Is there anything else to work on regarding this PR or just keep rebasing

jesperpedersen:
Merged.

